PR #3, opened 2021-10-03
Setting up Sequelize Models and Configurations

plausibly:
Setup the following models: Posts, Comments, and Users as well as their migration files.

This is what the models look like so far:

image
image
image

Also updated config file, it uses Heroku's environment variable DATABASE_URL to connect to the database.

### DIFF
--- a/nginx/conf.d/default.conf.template
+++ b/nginx/conf.d/default.conf.template
@@ -6,21 +6,13 @@ server {
         root   /usr/share/nginx/html;
         index  index.html index.htm;
     }
+    
+    location /api {
+        proxy_pass  http://127.0.0.1:8080;
+    }
 
     error_page   500 502 503 504  /50x.html;
     location = /50x.html {
         root   /usr/share/nginx/html;
     }
-}
-
-server {
-    listen $PORT;
-    server_name api.localhost;
-
-    location / {
-        proxy_pass  http://localhost:8080/v1/;
-        proxy_set_header Host $host;
-    }
-
-
 }
--- a/server/config/config.ts
+++ b/server/config/config.ts
@@ -1,0 +1,5 @@
+module.exports = {
+    DB_URL: process.env.DATABASE_URL || "postgres://vlynoaoa:ZPD-xj3wMzTFF4pKggWYU_dmZUeyYOJ9@fanny.db.elephantsql.com/vlynoaoa", // testing URL doesnt matter
+    DIALECT: "postgres"
+  
+};
--- a/server/config/config.ts
+++ b/server/config/config.ts
@@ -1,5 +1,4 @@
 module.exports = {
-    DB_URL: process.env.DATABASE_URL
-    DIALECT: "postgres"
-  
+    DB_URL:  process.env.DATABASE_URL, 
+    dialect: "postgres",
 };
--- a/server/config/config.ts
+++ b/server/config/config.ts
@@ -1,5 +1,5 @@
 module.exports = {
-    DB_URL: process.env.DATABASE_URL || "postgres://vlynoaoa:ZPD-xj3wMzTFF4pKggWYU_dmZUeyYOJ9@fanny.db.elephantsql.com/vlynoaoa", // testing URL doesnt matter
+    DB_URL: process.env.DATABASE_URL
     DIALECT: "postgres"
   
 };
--- a/server/migrations/20210925190654-create-user.ts
+++ b/server/migrations/20210925190654-create-user.ts
@@ -2,7 +2,7 @@
 module.exports = {
   up: async (queryInterface: any, Sequelize: any) => {
     await queryInterface.createTable("Users", {
-      uid: {
+      id: {
         allowNull: false,
         primaryKey: true,
         type: Sequelize.UUIDV4,
@@ -12,7 +12,6 @@ module.exports = {
         type: Sequelize.STRING(8),
         allowNull: false,
         unique: true
-
       },
       password: {
         type: Sequelize.STRING,
--- a/server/migrations/20210925190654-create-user.ts
+++ b/server/migrations/20210925190654-create-user.ts
@@ -1,0 +1,47 @@
+"use strict";
+module.exports = {
+  up: async (queryInterface: any, Sequelize: any) => {
+    await queryInterface.createTable("Users", {
+      uid: {
+        allowNull: false,
+        primaryKey: true,
+        type: Sequelize.UUIDV4,
+        unique: true
+      },
+      userName: {
+        type: Sequelize.STRING(8),
+        allowNull: false,
+        unique: true
+
+      },
+      password: {
+        type: Sequelize.STRING,
+        allowNUll: false
+      },
+      salt: {
+        type: Sequelize.STRING,
+      },
+      email: {
+        type: Sequelize.STRING,
+        unique: true
+      },
+      lastLogin: {
+        type: Sequelize.DATE,
+      },
+      karma: {
+        type: Sequelize.INTEGER,
+      },
+      createdAt: {
+        allowNull: false,
+        type: Sequelize.DATE,
+      },
+      updatedAt: {
+        allowNull: false,
+        type: Sequelize.DATE,
+      },
+    });
+  },
+  down: async (queryInterface: any, Sequelize: any) => {
+    await queryInterface.dropTable("Users");
+  },
+};
--- a/server/migrations/20210925191009-create-post.ts
+++ b/server/migrations/20210925191009-create-post.ts
@@ -1,0 +1,47 @@
+"use strict";
+module.exports = {
+  up: async (queryInterface: any, Sequelize: any) => {
+    await queryInterface.createTable("Posts", {
+      id: {
+        allowNull: false,
+        primaryKey: true,
+        type: Sequelize.STRING,
+      },
+      userId: {
+        type: Sequelize.STRING,
+      },
+      title: {
+        type: Sequelize.TEXT,
+      },
+      body: {
+        type: Sequelize.STRING,
+      },
+      thumbnail: {
+        type: Sequelize.STRING,
+      },
+      tags: {
+        type: Sequelize.STRING,
+      },
+      location: {
+        type: Sequelize.STRING,
+      },
+      capacity: {
+        type: Sequelize.INTEGER,
+      },
+      feedbackScore: {
+        type: Sequelize.INTEGER,
+      },
+      createdAt: {
+        allowNull: false,
+        type: Sequelize.DATE,
+      },
+      updatedAt: {
+        allowNull: false,
+        type: Sequelize.DATE,
+      },
+    });
+  },
+  down: async (queryInterface: any, Sequelize: any) => {
+    await queryInterface.dropTable("Posts");
+  },
+};
--- a/server/migrations/20210925191009-create-post.ts
+++ b/server/migrations/20210925191009-create-post.ts
@@ -7,9 +7,6 @@ module.exports = {
         primaryKey: true,
         type: Sequelize.STRING,
       },
-      userId: {
-        type: Sequelize.STRING,
-      },
       title: {
         type: Sequelize.TEXT,
       },
--- a/server/migrations/20211001211616-create-comment.ts
+++ b/server/migrations/20211001211616-create-comment.ts
@@ -1,0 +1,26 @@
+"use strict";
+module.exports = {
+  up: async (queryInterface: any, Sequelize: any) => {
+    await queryInterface.createTable("Comments", {
+      id: {
+        allowNull: false,
+        primaryKey: true,
+        type: Sequelize.UUID,
+      },
+      body: {
+        type: Sequelize.TEXT,
+      },
+      createdAt: {
+        allowNull: false,
+        type: Sequelize.DATE,
+      },
+      updatedAt: {
+        allowNull: false,
+        type: Sequelize.DATE,
+      },
+    });
+  },
+  down: async (queryInterface: any, Sequelize: any) => {
+    await queryInterface.dropTable("Comments");
+  },
+};
--- a/server/models/index.ts
+++ b/server/models/index.ts
@@ -1,14 +1,14 @@
 'use strict';
 
-const fs = require('fs');
-const path = require('path');
-const Sequelize = require('sequelize');
-const config = require('../config/config');
+import fs from 'fs';
+import path from 'path';
+import Sequelize from 'sequelize';
+const config = require( '../config/config');
 const basename = path.basename(__filename);
 const db: any = {};
 
-const sequelize = new Sequelize(config.DB_URL,  {
-  dialect: config.DIALECET,
+const sequelize = new Sequelize.Sequelize(config.DB_URL,  {
+  dialect: config.dialect,
   dialectOptions: {
     ssl:  { /* Required for our host */
       rejectUnauthorized: false,
--- a/server/models/index.ts
+++ b/server/models/index.ts
@@ -1,0 +1,38 @@
+'use strict';
+
+const fs = require('fs');
+const path = require('path');
+const Sequelize = require('sequelize');
+const config = require('../config/config');
+const basename = path.basename(__filename);
+const db: any = {};
+
+const sequelize = new Sequelize(config.DB_URL,  {
+  dialect: config.DIALECET,
+  dialectOptions: {
+    ssl:  { /* Required for our host */
+      rejectUnauthorized: false,
+    }
+  }
+});
+
+fs
+  .readdirSync(__dirname)
+  .filter((file: string) => {
+    return (file.indexOf('.') !== 0) && (file !== basename) && (file.slice(-3) === '.js');
+  })
+  .forEach((file: any) => {
+    const model = require(path.join(__dirname, file))(sequelize, Sequelize.DataTypes);
+    db[model.name] = model;
+  });
+
+Object.keys(db).forEach(modelName => {
+  if (db[modelName].associate) {
+    db[modelName].associate(db);
+  }
+});
+
+db.sequelize = sequelize;
+db.Sequelize = Sequelize;
+
+export default db;
--- a/server/models/user.ts
+++ b/server/models/user.ts
@@ -5,7 +5,7 @@ import { Model, UUIDV4, DataTypes } from "sequelize";
 interface UserAttributes {
   
   /* Login Specific */
-  uid: string;
+  id: string;
   userName: string;
   password: string; // hash
   salt: string;
@@ -20,22 +20,25 @@ interface UserAttributes {
 module.exports = (sequelize: any) => {
 
   class User extends Model<UserAttributes> implements UserAttributes {
-    uid!: string; // uuidv4
+    id!: string; // uuidv4
     userName!: string;
     password!: string;
     salt!: string;
     email!: string;
+
     lastLogin!: Date;
-    karma!: Number;
+    karma!: Number; // Should not be revealed to public
 
     static associate(model: any){
-      User.hasMany(model.Post);
+      User.hasMany(model.Post); 
+      User.hasMany(model.Comment);
+      /* Userid is stored inside of respective Post, and Comment */
     }
 
   };
 
    User.init({
-    uid :
+    id :
     {
       type: DataTypes.UUID,
       defaultValue: UUIDV4,
--- a/server/models/user.ts
+++ b/server/models/user.ts
@@ -1,0 +1,89 @@
+"use strict";
+import { Model, UUIDV4, DataTypes } from "sequelize";
+
+
+interface UserAttributes {
+  
+  /* Login Specific */
+  uid: string;
+  userName: string;
+  password: string; // hash
+  salt: string;
+  email: string;
+
+  /* Logs */
+  lastLogin: Date;
+  karma: Number;
+};
+
+
+module.exports = (sequelize: any) => {
+
+  class User extends Model<UserAttributes> implements UserAttributes {
+    uid!: string; // uuidv4
+    userName!: string;
+    password!: string;
+    salt!: string;
+    email!: string;
+    lastLogin!: Date;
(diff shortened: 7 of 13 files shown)


server/tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "es5", /* Set the JavaScript language version for emitted JavaScript and include compatible library declarations. */
+    "target": "es6", /* Set the JavaScript language version for emitted JavaScript and include compatible library declarations. */
```
danielduong8, line 3: I think this should be left as `es5`. This option is simply to tell TS what version of JS it wants to compile to. We can code in es6+ regardless, but having this set as `es5` allows the compiled JS to be backwards compatible.

server/yarn.lock
```diff
@@ -105,6 +105,11 @@ ansi-align@^3.0.0:
   dependencies:
     string-width "^3.0.0"
 
+ansi-regex@^2.1.1:
```
danielduong8, line 108: Consider updating this, to resolve the security error displayed on GitHub. If it keeps reinstalling itself, try to force it to be installed by including it in the `package.json`

server/package.json
```diff
@@ -13,7 +13,9 @@
     "cors": "^2.8.5",
     "dotenv": "^10.0.0",
     "express": "^4.17.1",
-    "pg": "^8.7.1"
+    "pg": "^8.7.1",
+    "sequelize": "^6.6.5",
+    "sequelize-cli": "^6.2.0"
```
danielduong8, line 18: I think if this is a CLI then it can be moved to devDependencies since we don't need it for Docker.

nginx/conf.d/default.conf.template
```diff
@@ -6,21 +6,13 @@ server {
         root   /usr/share/nginx/html;
         index  index.html index.htm;
     }
+    
+    location /api {
+        proxy_pass  http://127.0.0.1:8080;
```
danielduong8, line 11: Changing this to `http://localhost:8080` might be better for clarity and consistency if we are going to use `localhost` in general.

server/models/user.ts
```diff
@@ -0,0 +1,92 @@
+"use strict";
+import { Model, UUIDV4, DataTypes } from "sequelize";
+
+
+interface UserAttributes {
```
danielduong8, line 5: Not sure if these interfaces are useful unless we are going to be using them to enforce attributes for other Classes that we define. If we are just using a single interface for a single class, this may be redundant.

danielduong8 (commented):
Try to have more detailed commit messages like Andy mentioned to me last time as well. Nice work though, overall!

server/yarn.lock
plausibly: Doing yarn install keeps bringing back the older versions of ansi in the yarn.lock, I think one of our other dependencies is relying on an older ansi-regex

server/tsconfig.json
plausibly: Sequelize seems to break if I switch this to es5. I will see if I can fix this

plausibly: Something broke while I was testing it, that's why I switched to ES6. I will switch it back to ES5 since it seems to work fine now.

server/migrations/20210925190654-create-user.ts
```diff
@@ -0,0 +1,46 @@
...
+      },
+      salt: {
+        type: Sequelize.STRING,
+      },
+      email: {
+        type: Sequelize.STRING,
+        unique: true
+      },
+      lastLogin: {
+        type: Sequelize.DATE,
```
andy9775, line 28: Can we add a hook here to automatically update this after an update happens?

server/migrations/20210925191009-create-post.ts
```diff
@@ -0,0 +1,44 @@
...
+      id: {
+        allowNull: false,
+        primaryKey: true,
+        type: Sequelize.STRING,
+      },
+      title: {
+        type: Sequelize.TEXT,
+      },
+      body: {
+        type: Sequelize.STRING,
```
andy9775, line 14: `STRING` defaults to 255 characters FYI. Is this enough?

server/migrations/20210925191009-create-post.ts
```diff
@@ -0,0 +1,44 @@
...
+      },
+      thumbnail: {
+        type: Sequelize.STRING,
+      },
+      tags: {
+        type: Sequelize.STRING,
+      },
+      location: {
```
andy9775, line 22: City? Country? 

It seems like the intention here is a front end field "Location". Lets be more specific what this means. Lets also figure out if we need City/Country or just one?

server/migrations/20210925191009-create-post.ts
```diff
@@ -0,0 +1,44 @@
...
+        type: Sequelize.STRING,
+      },
+      capacity: {
```
andy9775, line 25: What's capacity?

server/migrations/20210925191009-create-post.ts
```diff
@@ -0,0 +1,44 @@
...
+        type: Sequelize.INTEGER,
+      },
+      feedbackScore: {
```
andy9775, line 28: feedbackScore?

server/server.ts
```diff
@@ -1,22 +1,21 @@
...
 
 const app = express();
 const port = 8080;
-
+ 
 app.use(express.json());
 app.use(express.urlencoded({ extended: true }));
 
-app.get('/v1/hello', (req: express.Request, res: express.Response) => {
+app.get("/test", (req: express.Request, res: express.Response) => {
```
andy9775, line 13:
```suggestion
app.get('/v1/hello', (req: express.Request, res: express.Response) => {
```

Lets keep minimal changes

server/models/index.ts
```diff
@@ -0,0 +1,38 @@
...
+Object.keys(db).forEach(modelName => {
+  if (db[modelName].associate) {
+    db[modelName].associate(db);
+  }
+});
+
+db.sequelize = sequelize;
+db.Sequelize = Sequelize;
+
+export default db;
```
andy9775, line 38: Is there a way to autogenerate some code? 

See https://github.com/sequelize/sequelize-auto or similar tools

server/migrations/20210925191009-create-post.ts
plausibly: A posts score is updated based on interactions, and if it's reported. We want to automatically remove posts that are often reported, so we can have a score that goes down on reports.

server/models/user.ts
plausibly: The sequelize docs says to define your own models like this

server/migrations/20210925190654-create-user.ts
```diff
@@ -0,0 +1,46 @@
...
+      },
+      karma: {
+        type: Sequelize.INTEGER,
+      },
+      createdAt: {
```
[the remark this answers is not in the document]
plausibly: Sequelize already sets the createdAt and modifiedAt by default without needing to explicitly state it

danielduong8:
Just as a note, I don't think we will need the `salt` key in the database since argon2 will add the `salt` before hashing.

andy9775:
Is the salt stored alongside the password hash?

server/models/index.ts
```diff
@@ -0,0 +1,38 @@
+'use strict';
+
+import fs from 'fs';
+import path from 'path';
+import Sequelize from 'sequelize';
+const config = require( '../config/config');
+const basename = path.basename(__filename);
+const db: any = {};
+
+const sequelize = new Sequelize.Sequelize(config.DB_URL,  {
```
andy9775, line 10: Can you update this to include
```
const sequelize = new Sequelize('sqlite::memory:')
```
if we're in test? You should be able to check for `ci` to be true in the environment.

server/migrations/20210925190654-create-user.ts
```diff
@@ -0,0 +1,46 @@
...
+      },
+      password: {
+        type: Sequelize.STRING,
+        allowNUll: false
+      },
+      salt: {
+        type: Sequelize.STRING,
+      },
+      email: {
+        type: Sequelize.STRING,
```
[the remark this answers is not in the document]
plausibly: Ok

server/migrations/20210925191009-create-post.ts
```diff
@@ -0,0 +1,44 @@
+"use strict";
+module.exports = {
+  up: async (queryInterface: any, Sequelize: any) => {
+    await queryInterface.createTable("Posts", {
+      id: {
```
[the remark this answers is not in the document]
plausibly: I will fix this, it should've been UUID.

server/migrations/20210925191009-create-post.ts
plausibly: The location needs to work with google maps API which is something we'll implement later.. for now it is just a placeholder. We will likely store coordinates.

danielduong8:
> Is the salt stored alongside the password hash?

I believe so, it looks something like this?

`$argon2i$v=19$m=65536,t=3,p=1$YOtX2//7NoD/owm8RZ8llw==$fPn4sPgkFAuBJo3M3UzcGss3dJysxLJdPdvojRF20ZE=`

Source:
https://security.stackexchange.com/questions/183083/is-it-safe-wise-to-store-a-salt-in-the-same-field-as-the-hashed-password

andy9775:
Sounds good. As long as its done automatically

server/models/index.ts
plausibly: Done

server/config/config.ts
```diff
@@ -0,0 +1,4 @@
+module.exports = {
```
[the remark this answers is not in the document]
plausibly: Unfortunately we can't do this on the free tier :(

server/migrations/20210925190654-create-user.ts
plausibly: I think this can be done after the login feature is added. I'm not sure how to add a hook without having defined what it means to be logged in.

server/migrations/20210925191009-create-post.ts
```diff
@@ -0,0 +1,44 @@
...
+        type: Sequelize.TEXT,
+      },
+      body: {
+        type: Sequelize.STRING,
+      },
+      thumbnail: {
+        type: Sequelize.STRING,
+      },
+      tags: {
+        type: Sequelize.STRING,
```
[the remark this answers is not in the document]
plausibly: Updated. This can be done with associations on the database

server/migrations/20210925190654-create-user.ts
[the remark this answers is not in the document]
andy9775: Sequelize adds a field for us, or automatically sets the value of `createdAt` when the model is created?

server/tsconfig.json
andy9775: Changing back to es5 still? Or are we keeping it?

server/models/user.ts
```diff
@@ -0,0 +1,86 @@
...
+      type: DataTypes.STRING,
+      allowNull: false
+    },   
+    email:
+    {
+      type: DataTypes.STRING(64),
+      allowNull: false,
+      unique: true,
+      validate : {
+         is: [".[a-zA-z\-\_\.0-9]*@[a-zA-z\-\_\.0-9]*utoronto.ca"],  // only utor emails
```
andy9775, line 65: Lets write some tests for this.

server/models/tags.ts
```diff
@@ -0,0 +1,48 @@
...
+      unique: true
+      },
+
+      text: {
+        type: DataTypes.STRING(24),
+        allowNull: false,
+        primaryKey: true,
+        unique: true,
+        validate: {
+            is: ["[A-Z][a-z]*"] // only alphabetical tags, no other characters
```
andy9775, line 38: `ClassOf2022` is not a valid tag?

server/models/tags.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  Tag.init(
+    {
+      id: {
+      type: DataTypes.UUID,
+      defaultValue: UUIDV4,
+      allowNull: false,
+      primaryKey: true,
+      unique: true
+      },
```
andy9775, line 30: I think the `text` should be ok enough to use as a primary key.

server/models/index.ts
```diff
@@ -0,0 +1,48 @@
...
+const basename = path.basename(__filename);
+const db: any = {};
+
+let sequelize: Sequelize.Sequelize;
+
+if (process.env.CI){
+  /* For CI testing */
+  sequelize = new Sequelize.Sequelize('sqlite::memory');
+}
+else{
```
andy9775, line 16:
```suggestion
} else {
```

server/models/index.ts
```diff
@@ -0,0 +1,48 @@
...
+import fs from 'fs';
+import path from 'path';
+import Sequelize from 'sequelize';
+const config = require( '../config/config');
+const basename = path.basename(__filename);
+const db: any = {};
+
+let sequelize: Sequelize.Sequelize;
+
+if (process.env.CI){
```
andy9775, line 12:
```
if ('false') {...}
```
Environment variables tend to be passed in as strings. You can lowercase and compare.

server/migrations/20211001211616-create-comment.ts
```diff
@@ -0,0 +1,27 @@
...
+    await queryInterface.createTable("PostComments", {
+      id: {
+        allowNull: false,
+        primaryKey: true,
+        type: Sequelize.UUID,
+        unique: true
+      },
+      body: {
+        type: Sequelize.TEXT,
+      },
```
andy9775, line 13:
```suggestion
      body: {
        type: Sequelize.TEXT,
        allowNull: false,
      },
```
We can also perform a non-empty check using hooks or a regex.

server/migrations/20210925191009-create-post.ts
andy9775: Fair enough. It may not make sense to use coordinates though since they're not human readable. You may want to store the full address in the db.

nginx/conf.d/default.conf.template
```diff
@@ -6,21 +6,13 @@ server {
         root   /usr/share/nginx/html;
         index  index.html index.htm;
     }
+    
+    location /api {
+        proxy_pass  http://localhost:8080;
+    }
```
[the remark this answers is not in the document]
andy9775: Sorry, meant a separate PR.

server/tsconfig.json
plausibly: We will do ES6, since that is what we switched to in the client directory and what Daniel did on his PR

server/migrations/20210925190654-create-user.ts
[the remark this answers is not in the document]
plausibly: Both, "By default, Sequelize will add the attributes createdAt and updatedAt to your model so you will be able to know when the database entry went into the db and when it was updated last." https://sequelize.org/v5/manual/models-definition.html

I also tested this by creating a user and it automatically sets those fields to the time that the item was added to the db

server/models/user.ts
plausibly: Sure. I will write database tests and make a separate PR after

andy9775: Can you do it in this one? It's better to keep the tests and the code they test on a single pr.

server/migrations/20210925190654-create-user.ts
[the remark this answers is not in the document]
andy9775: That's what I thought. You should be able to get away with not defining them explicitly if Sequelize automatically adds it for us. So you can remove all `createdAt` and `updatedAt` fields.

server/models/user.ts
plausibly: Ok

nginx/conf.d/default.conf.template
```diff
@@ -11,16 +11,4 @@ server {
...
-server {
-    listen $PORT;
-    server_name api.localhost;
-
-    location / {
-        proxy_pass  http://localhost:8080/v1/;
-        proxy_set_header Host $host;
-    }
-
-
```
andy9775, line 25: Don't add or remove anything from this file leave it as it was. A pr should be focused and changing nginx config in a pr which focuses on models and database related stuff violates that.

server/migrations/20210925190654-create-user.ts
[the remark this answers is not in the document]
andy9775: got it

plausibly (commented):
Reviewed Changes

server/models/tests/post.test.ts
```diff
@@ -0,0 +1,41 @@
...
+            expect(posted.title).toEqual("This is a new post!")
+        
+        });
+    });
+
+    describe('Post Validation', () => {
+        test("Attempt to make posts with invalid or empty author", async() => {
+            await expect(makePost(null, "asd", "aaaaaaaaa")).rejects.toThrowError("notNull Violation"); // NULL author
+        
+            await expect(makePost("123123", "asdddd", "Woooo attempting new post abcdefghhi!!")).rejects.toThrowError("SQLITE_CONSTRAINT"); // Non-existant uid
```
andy9775, line 28: Can you re-format this bit of code?

server/models/tests/testHelpers.ts
```diff
@@ -0,0 +1,45 @@
...
+
+        }).catch( (err: Error) => { throw err });
+
+        return testUser;
+
+}
+
+/* Ensure that the database is synchronized properly */
+export function dbSync() { 
+        test("Test model sync", async () => {
```
andy9775, line 39: This shouldn't be a test. You should raise an error if this is invalid.

andy9775 (changes requested):

server/migrations/20210925190654-create-user.ts
```diff
@@ -0,0 +1,70 @@
...
+        defaultValue: UUIDV4,
+        allowNull: false,
+        primaryKey: true,
+        unique: true
+      },
+      firstName: {
+        type: Sequelize.STRING,
+        allowNull: false,
+        validate: {
+          isAlpha: true
```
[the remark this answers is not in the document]
plausibly: isAlpha means is alphabetical in sequelize

server/migrations/20210925190654-create-user.ts
```diff
@@ -0,0 +1,70 @@
...
+      email: {
+        type: Sequelize.STRING(64),
+        allowNull: false,
+        unique: true, /* In the controller, lets force email to lowercase */
+        validate : {
+           isEmail: true, /* Check for valid email format */
+           is: [".*@(mail\.||alum\.||^$)utoronto.ca"],  /* Check for utoronto domain */
+        }
+      },
+      confirmed: {
```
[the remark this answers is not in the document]
plausibly: This might not be necessary for JWT. The tokens could be stored maybe as a cookie.. JWT tokens can be validated in the backend to see what data it contains. The expiry date is also stored in the token itself.

So if a user makes an API call to create a post, we could send their login token alongside the request. The token can be checked in the back-end to see if it's valid and matches their id, then we proceed to make the post..

server/models/tests/post.test.ts
plausibly: Sure. My prettier wasn't working for some reason, I will fix it and autoformat the code.

server/migrations/20210925190654-create-user.ts
```diff
@@ -0,0 +1,70 @@
...
+        unique: true
+      },    
+      password: {
+        type: Sequelize.STRING,
+        allowNull: false
+      },   
+      email: {
+        type: Sequelize.STRING(64),
+        allowNull: false,
+        unique: true, /* In the controller, lets force email to lowercase */
```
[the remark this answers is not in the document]
plausibly: Sure

plausibly (commented):
Reviewed requested changes / comments

danielduong8:
I think it would be better to put the formatting fixes in a separate commit in the future. Otherwise, I think it looks good to me!

server/config/config.ts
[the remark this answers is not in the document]
andy9775: Lets add a TODO since we now have funds.

server/migrations/20210925190654-create-user.ts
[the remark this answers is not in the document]
andy9775: When you create the user account and send a confirmation email you want to use a token to validate the user. Otherwise anyone can validate the account. Further, when a user issues a password reset request you use the token to verify that the user received and clicked the correct link. The expiry prevents stale password resets. Finally, if user_a changes their password we would clear the token. Thus, user_b wouldn't be able to reuse the password reset link via a replay attack. Now, you could include the user email in the password reset link, but then anyone can just go to example.com?email=foo@bar.com and reset the users password.

This isn't an API token. It's related to account creation and password resets.

server/models/user.ts
```diff
@@ -0,0 +1,114 @@
...
+  password: string; // hash
+  email: string;
+  confirmed: boolean;
+
+  /* Logs */
+  lastLogin: Date;
+  karma: Number;
+}
+
+/* Useful for typescript definitions of Users */
```
andy9775, line 18:
```suggestion

```

andy9775 (changes requested):
Major change is the token. Feel free to reach out on discord if you're not clear about the purpose of the token. We can get away with just having the single field as well.

server/migrations/20210925190654-create-user.ts
```diff
@@ -56,6 +56,12 @@ module.exports = {
         type: Sequelize.INTEGER,
         defaultValue: 0,
       },
+      token: {
+        type: Sequelize.STRING,
+      },
+      tokenExpires: {
+        type: Sequelize.DATE,
+      },
```
andy9775, line 64:
```suggestion
      confirmationToken: {
        type: Sequelize.STRING,
      },
      confirmationTokenExpires: {
        type: Sequelize.DATE,
      },
```

PR #4, opened 2021-10-03
Created a login and sign up component 

gurvirsasan:
Worked in the client directory on the front end.
Created a login and sign up component for the landing page.
Also, placed them in a container so that on can toggle the tabs to switch from login to sign up and vice versa.


client/src/components/SignUp.tsx
```diff
@@ -0,0 +1,118 @@
...
+          type="password"
+        />
+
+        <TextField
+          label="Phone Number"
+          placeholder="Enter Password"
+          onChange={(e) => setPhNum(e.target.value)}
+          fullWidth
+          required
+          type="password"
```
andy9775, line 75:
```suggestion
          type="tel"
```

client/src/components/SignUp.tsx
```diff
@@ -0,0 +1,118 @@
...
+          placeholder="youremail@mail.utoronto.ca"
+          onChange={(e) => setEmail(e.target.value)}
+          fullWidth
+          required
+          type="password"
+        />
+
+        <TextField
+          label="Phone Number"
+          placeholder="Enter Password"
```
andy9775, line 71:
```suggestion
          placeholder="Enter Phone Number"
```

client/src/components/SignUp.tsx
```diff
@@ -0,0 +1,118 @@
...
+          required
+        />
+
+        <TextField
+          label="Email"
+          placeholder="youremail@mail.utoronto.ca"
+          onChange={(e) => setEmail(e.target.value)}
+          fullWidth
+          required
+          type="password"
```
andy9775, line 66:
```suggestion
          type="email"
```

client/src/components/SignUp.tsx
```diff
@@ -0,0 +1,118 @@
...
+      </Grid>
+
+      <Box textAlign="center">
+        <h1>Sign Up</h1>
+      </Box>
+      <form noValidate autoComplete="off" onSubmit={handleSubmit}>
+        <TextField
+          label="Name"
+          placeholder="FirstName LastName"
+          onChange={(e) => setUsername(e.target.value)}
```
andy9775, line 55: Is the first/last name the username? Shouldn't username be something like `gurvirsasan`, or `andy9775`?

client/src/App.css
```diff
@@ -1,5 +1,11 @@
 .App {
```
andy9775, line 1: Keep in mind that this css is applied to the `App` component - it is application wide.

client/src/components/Login.tsx
```diff
@@ -0,0 +1,88 @@
...
+    padding: 18,
+    borderRadius: "20px",
+    height: 570,
+    width: 350,
+  };
+
+  const avatarStyle = {
+    backgroundColor: "#f4003d",
+    width: 50,
+    height: 50,
```
andy9775, line 17: Ditto

andy9775 (commented):
The content of `client/src/containers/index.tsx` should be its own file/component. `auth.tsx` perhaps? Index.tsx should export the components which need to be public and nothing more. Then app.js should setup the routing of each page.

Otherwise good job! Good organization and structure, and thanks for keeping the commit short 😄

client/tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "es5",
+    "target": "es6",
```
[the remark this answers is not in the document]
plausibly: Does ES5 not work with the react code? If ES6 is required for the front-end to function then we can keep it this way.

andy9775:
Also, one more comment. Lets add some tests. We should assert the default component and assert that toggling changes the UI correctly.

client/src/App.css
```diff
@@ -1,5 +1,11 @@
...
   text-align: center;
+  background-image: url("https://media.gettyimages.com/videos/loopable-color-gradient-background-animation-video-id1182636595?s=640x640");
```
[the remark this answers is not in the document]
gurvirsasan: No, I haven't checked the licensing; I just used a basic image to have some visual. Would you recommend removing it for now or to replace it (given licensing issues) in future?

gurvirsasan:
In terms of structuring, I'm going to take the contents out of  client/src/containers/index.tsx and make it into its own file under client/src/components.

I'm a little confused about the purpose of index file. What does it mean by "exporting components which need to be public"

Also, should this index file be under its own folder? like client/src/containers/index.tsx? or should it be under client/src/components/index.tsx. In the latter case, I think it won't be beneficial as we won't be able to use any more components and an "index" file exporting their components that need to be public.

andy9775:
The current index.tsx file should stay where it is. It should re-export any public components. Then inside of app.js you can do 

`import {A,B,C} from 'containers'`

instead of 

`import A from 'containers/A'`
`import B from 'containers/B'`
`import C from 'containers/C'`.

The current fils should re-export any container components. 

Having the index file re-export is optional. Another option is something like this: https://github.com/demoreno/react-structure/blob/master/src/routes.js. Notice though how each container is its own file. With the current structure, having the login page logic inside of 'containers/index.tsx' implies that the login page is the main component for the containers directory. It should be 'containers/login.tsx`.

plausibly (commented):
Good work

client/src/components/SignUp.tsx
```diff
@@ -0,0 +1,241 @@
+import React, { useState } from "react";
+
+import { Grid, Paper, Avatar, TextField, Button, Box } from "@material-ui/core";
+import CircleOutlined from "@material-ui/icons/AddCircleOutlineOutlined";
+import useMediaQuery from "@material-ui/core/useMediaQuery";
+
+interface SignUpProps {
+  handleChange: Function;
```
plausibly, line 8: Any way to do this without defining the interface? Not sure if we should have one if it is only used in one spot

client/src/containers/AuthContainer.test.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+    // toggle to log in from in-form button
+    screen.getByTestId("GoToLogIn").click();
+    const checkLogIn = screen.getByTestId("LogInTab");
+    expect(checkLogIn).toBeInTheDocument();
+  });
+});
+
+describe("verifying valid input for sigup page", () => {
+  it("only allow valid uToronto email addresse", () => {
```
plausibly, line 68: address*

client/src/containers/index.tsx
```diff
@@ -0,0 +1,3 @@
+import AuthContainer from "./AuthContainer";
+
```
plausibly, line 2: Seems useless.. maybe we should use this file for a welcome screen in a later PR.

plausibly (approved):
Looks good.. just some minor comments

client/src/containers/index.tsx
gurvirsasan: Just using index.tsx as a centralized place for containers and components. We can definitely change this as we proceed.

client/src/containers/AuthContainer.test.tsx
```diff
@@ -65,7 +65,7 @@ describe("test Auth Pages", () => {
 });
 
 describe("verifying valid input for sigup page", () => {
```
plausibly, line 67: sigup

PR #7, opened 2021-10-09
Integrate GitHub Actions for CI

danielduong8:
The client and server directories have `.test` template test files that you can use to structure your tests if needed. If you need help with some of the jest syntax or structure you can take a look at this link: https://jestjs.io/docs/getting-started .

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -1,6 +1,6 @@
 name: Tests CI
 
-on: [push, pull_request]
+on: [push]
 
 jobs:
   test:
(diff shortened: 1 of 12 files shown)


.github/workflows/node.js.yml
```diff
@@ -0,0 +1,48 @@
...
+        working-directory: ${{ matrix.dir }}
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v2
+      with:
+        node-version: ${{ matrix.node-version }}
+    - run: yarn install
+    - run: yarn run build
```
andy9775, line 26: No need to build

PR #9, opened 2021-10-17
Resolve typing and errors related to typing

danielduong8:
- Remove config.ts
config.ts can be replaced with a .env file together along with the
dotenv package.

- Fix imports for models and migrations
Several files do not import `Sequelize` from "sequelize".

- Replace most `any` types with correct types

- Restructure logic for models/index.ts and use correct db url
Move SQLite case to else block to ensure `sequelize` is always
defined. Also, use correct DATABASE_URL environment variable which
is defined by Heroku. Finally, add protocol and ssl require
options to support Heroku.

- Add interface for optional creation attributes
Tests which use `db.user` do not have correct types which
allows models to be created with missing attributes. After fixing
types, user.ts and post.ts should be updated to allow certain
attributes to be optional during creation.

- Add missing interface declaration in posts.ts
UserId is not defined in the PostAttributes interface.

- Validate UserId for empty strings in posts.ts

- Change `sequelize` to `Sequelize` for `.fn` & `.col` in user.ts
`.fn` and `.col` are static functions and therefore need to be
called with `Sequelize.`.

- Correct typing and tests for model/tests files

### DIFF
--- a/server/migrations/20210925190654-create-user.ts
+++ b/server/migrations/20210925190654-create-user.ts
@@ -1,31 +1,31 @@
-import { UUIDV4 } from "sequelize";
+import { Sequelize, QueryInterface, DataTypes, UUIDV4 } from "sequelize";
 
 module.exports = {
-  up: async (queryInterface: any, Sequelize: any) => {
+  up: async (queryInterface: QueryInterface, Sequelize: Sequelize) => {
     await queryInterface.createTable("Users", {
       id: {
-        type: Sequelize.UUID,
+        type: DataTypes.UUID,
         defaultValue: UUIDV4,
         allowNull: false,
         primaryKey: true,
         unique: true,
       },
       firstName: {
-        type: Sequelize.STRING,
+        type: DataTypes.STRING,
         allowNull: false,
         validate: {
           isAlpha: true,
         },
       },
       lastName: {
-        type: Sequelize.STRING,
+        type: DataTypes.STRING,
         allowNull: false,
         validate: {
           isAlpha: true,
         },
       },
       userName: {
-        type: Sequelize.STRING(32),
+        type: DataTypes.STRING(32),
         allowNull: false,
         unique: true,
         validate: {
@@ -33,11 +33,11 @@ module.exports = {
         },
       },
       password: {
-        type: Sequelize.STRING,
+        type: DataTypes.STRING,
         allowNull: false,
       },
       email: {
-        type: Sequelize.STRING(64),
+        type: DataTypes.STRING(64),
         allowNull: false,
         unique: true /* In the controller, lets force email to lowercase */,
         validate: {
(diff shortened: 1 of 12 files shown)


server/models/post.ts
```diff
@@ -68,6 +73,13 @@ module.exports = (sequelize: Sequelize.Sequelize) => {
         type: DataTypes.INTEGER,
         defaultValue: 1,
       },
+      UserId: {
+        type: DataTypes.STRING,
```
plausibly, line 77: UserId is type DataTypes.UUID

server/migrations/20210925190654-create-user.ts
```diff
@@ -1,43 +1,43 @@
-import { UUIDV4 } from "sequelize";
+import { Sequelize, QueryInterface, DataTypes, UUIDV4 } from "sequelize";
 
 module.exports = {
-  up: async (queryInterface: any, Sequelize: any) => {
+  up: async (queryInterface: QueryInterface, Sequelize: Sequelize) => {
```
plausibly, line 4: Since you defined Userid field in the model, you should add it here too

server/models/user.ts
```diff
@@ -38,7 +41,7 @@ export class User extends Model<UserAttributes> implements UserAttributes {
   }
```
plausibly, line 41: Since you defined UserId in the Post model, for the association you should do:

`User.hasMany(model.Post, { foreignKey: {name: "UserId", allowNull: false } });`

plausibly (changes requested):
- Change UserId to type UUID
-  Add UserId to migration (should match the model file)
- Declare foreignKey name in User.ts

plausibly (commented):
client changes lgmt

server/migrations/20210925190654-create-user.ts
danielduong8: Do you mean in create-post.ts, not create-user.ts?

plausibly: yes, my bad

danielduong8:
Ready for review~

plausibly (approved):

server/models/index.ts
```diff
@@ -1,24 +1,32 @@
...
-if (process.env.CI && process.env.CI.toLowerCase() === "true") {
-  /* For CI testing */
-  sequelize = new Sequelize.Sequelize("sqlite::memory");
-} else {
-  sequelize = new Sequelize.Sequelize(config.DB_URL, {
-    dialect: config.dialect,
+if (
+  (!process.env.CI || process.env.CI.toLowerCase() === "false") &&
+  process.env.DATABASE_URL
+) {
```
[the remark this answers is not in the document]
danielduong8: What should happen in the case where both conditions are false?
`
if (process.env.CI && process.env.CI.toLowerCase() === "false") {
  /* For CI testing */
  sequelize = new Sequelize("sqlite::memory");
} else if (process.env.DATABASE_URL) {
  sequelize = new Sequelize(process.env.DATABASE_URL, {
    dialect: "postgres",
    protocol: "postgres",
    dialectOptions: {
      ssl: {
        require: true,
        /* Required for our host */ rejectUnauthorized: false,
      },
    },
  });
} else {
  ?
}
`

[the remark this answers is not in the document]
danielduong8: I was thinking if `process.env.DATABASE_URL` is missing, we simply use sqlite, like so?
```
if (process.env.CI && process.env.CI.toLowerCase() === "true" || !process.env.DATABASE_URL) {
  /* For CI testing */
  sequelize = new Sequelize("sqlite::memory");
} else {
  sequelize = new Sequelize(process.env.DATABASE_URL, {
    dialect: "postgres",
    protocol: "postgres",
    dialectOptions: {
      ssl: {
        require: true,
        /* Required for our host */ rejectUnauthorized: false,
      },
    },
  });
}
```

PR #10, opened 2021-10-20
Add routes and UserController + Auth

danielduong8:
- Add sqlite to Docker image for local testing
Update README.md accordingly.

- Add auth middleware

- Add signIn and createUser methods under UserController

- Add route handlers for UserController methods
Store jwts in cookies.

- Add signup, signin, and signout functionality

- Add tests for routes and UserController methods

- Add router to /api/v1/user endpoint
Remove placeholder route in server.ts.

- Remove unused argument in models/index.ts

- Edit README.md to include setting env CI to false

==============================

Auth middleware should be used as follows:
```
import auth from "../middleware/auth";

router.get("/some_endpoint", auth, controllerMethodThatNeedsAuthorization);
```


server/@types/index.d.ts
```diff
@@ -0,0 +1,5 @@
+declare namespace Express {
+  interface Request {
+    userId: string;
```
plausibly, line 3: Is this necessary? You should be able to `import { Request } from "express"` and use userid as a param

server/controllers/v1/user.ts
```diff
@@ -0,0 +1,92 @@
...
+  static signIn = async (req: Request, res: Response) => {
+    const { userName, password } = req.body;
+
+    try {
+      const oldUser = await UserController.UserModel.findOne({
+        where: {
+          userName: userName,
+        },
+      });
+      if (!oldUser)
```
plausibly, line 26: Does this work if an error occurs in the `try` case? Might need a `catch` statement

server/controllers/v1/user.ts
```diff
@@ -0,0 +1,92 @@
...
+  static createUser = async (req: Request, res: Response) => {
+    const { email, userName, password, firstName, lastName } = req.body;
+
+    try {
+      const oldUser = await UserController.UserModel.findOne({
+        where: {
+          userName: userName,
+        },
+      });
+      if (oldUser)
```
plausibly, line 62: same thing, I think you should use `catch` to handle db errors

server/controllers/v1/user.ts
```diff
@@ -0,0 +1,92 @@
...
+        return res.status(400).json({ message: "User already exists" });
+
+      const hashedPassword = await argon2.hash(password, {
+        type: argon2.argon2id,
+      });
+
+      const result: User = await UserController.UserModel.create({
```
plausibly, line 69: This should be in a `try catch` statement, if the user has not properly formatted their information the database can throw errors (such as `ValidationError`)

danielduong8: The catch statement can be found below at line 86

server/controllers/v1/user.ts
danielduong8: The catch statement can be found below at line 86

server/controllers/v1/user.ts
```diff
@@ -9,13 +11,117 @@ export enum TOKEN_TYPE {
...
+        password: hashedPassword,
+        firstName: firstName,
+        lastName: lastName,
+      });
+
+      const token = jwt.sign(
+        { userName: result.userName, id: result.id },
+        this.secret,
+        { expiresIn: "1h" }
+      );
```
andy9775, line 114: Creating a new user record shouldn't sign the user in (create a token). The user needs to confirm their account.

server/controllers/v1/user.ts
```diff
@@ -9,13 +11,117 @@ export enum TOKEN_TYPE {
...
+
+      const token = jwt.sign(
+        { username: oldUser.userName, id: oldUser.id },
+        this.secret,
+        { expiresIn: "1h" }
+      );
+
+      await this.updateLastLogin(oldUser);
+
+      return { status: 200, data: { result: oldUser, token: token } };
```
andy9775, line 69: Token should not be sent as part of the JSON response - cookie only.

as an implementation reference: https://cheatcode.co/tutorials/how-to-implement-secure-httponly-cookies-in-node-js-with-express

server/controllers/v1/user.ts
```diff
@@ -9,13 +11,117 @@ export enum TOKEN_TYPE {
 
 export default class UserController {
   protected userRepo: typeof User;
+  secret: string;
   protected emailService: EmailService;
 
   constructor(model: typeof User, emailService: EmailService) {
     this.userRepo = model;
+    this.secret = process.env.SECRET ? process.env.SECRET : "secret";
```
andy9775, line 19: Not secure to use a default secret - we should error out if one hasn't been provided. You could image a deploy procedure which omits the secret and therefore it always defaults to "secret". In cases where the ENV variable is set, but not passed in would hide this issue - you may think it's being passed in, but it's not. Further, if this is the case, when you rotate the secret, it never is actually rotated since it's always the string "secret"

server/controllers/v1/user.ts
```diff
@@ -9,13 +11,117 @@ export enum TOKEN_TYPE {
...
+   * Generate auth token, update last login date, and send confirmation email.
+   * @returns status code and data payload
+   */
+  async createUser(
+    email: string,
+    userName: string,
+    password: string,
+    firstName: string,
+    lastName: string
+  ): Promise<{ status: number; data: any }> {
```
andy9775, line 87: We should be able to specify a type for the response since it follows a specific structure.

andy9775 (changes requested):
Just some comments after taking a first pass

server/controllers/v1/user.ts
danielduong8: I've moved both this and updateLastLogin to the confirmEmail method.

server/controllers/v1/user.ts
danielduong8: The UserController has no access to `res` so how should I go about calling `res.cookie` without passing it back to routes?

server/controllers/v1/user.ts
danielduong8: Fixed, an error should now be thrown when secret is undefined.

server/controllers/v1/user.ts
danielduong8: I specified the type as:
`Promise<{ status: number; data: { result?: User, token?: string, message?: string } }>`
since it may only return message on failure.

server/controllers/v1/tests/user.test.ts
```diff
@@ -32,7 +45,105 @@ beforeEach(async () => {
   fakeSendResetEmail.mockClear();
 });
 
-describe("Test v1 - User Controller", () => {
+describe("v1 - User Controller", () => {
+  beforeAll(async () => {
+    createUserDate = new Date();
+    createUserResponse = await uContr.createUser(
```
plausibly, line 51: you could use the makeUser helper from testHelpers

server/routes/v1/tests/user.test.ts
```diff
@@ -0,0 +1,88 @@
...
+function mockResponse() {
+  const res: any = {};
+  res.status = jest.fn().mockReturnValue(res);
+  res.json = jest.fn().mockReturnValue(res);
+  res.cookie = jest.fn().mockReturnValue(res);
+  return res;
+}
+
+beforeAll(async () => {
+  await db.sequelize.sync({ force: true });
```
plausibly, line 19: There is a helper function that does this, check the testHelpers file in user models test

server/routes/v1/tests/user.test.ts
```diff
@@ -0,0 +1,88 @@
...
+});
+
+afterAll(async () => {
+  await db.sequelize.sync({ force: true });
```
plausibly, line 23: same as above comment

plausibly (commented):
Looks good, see comment above

server/controllers/v1/user.ts
andy9775: Return [token, data]

server/controllers/v1/user.ts
andy9775: Confirming email shouldn't create a token either. Only sign in should create a token. Signout should delete the token

server/server.test.ts
```diff
@@ -1,8 +1,8 @@
...
-      });
+describe("Unit...", () => {
+  describe("Method...", () => {
+    test("Scenario...", () => {
+      expect(true).toBeTruthy();
     });
-  });
\ No newline at end of file
+  });
+});
```
andy9775: If the test doesn't test any logic let's remove it

server/routes/v1/user.ts
```diff
@@ -0,0 +1,60 @@
...
+
+  if (response.data.token) {
+    res.cookie("token", JSON.stringify(response.data.token), {
+      httpOnly: true,
+      expires: moment().add(1, "hour").toDate(),
+      signed: true,
+    });
+  }
+
+  res.status(response.status).json(response.data);
```
danielduong8, line 52: I see, yeah I'm passing all of data including the token here...
Can I simply change this line to be this instead?
`res.status(response.status).json(response.data.result);`

server/controllers/v1/user.ts
andy9775: Or, better yet generate the token in the routes file since the token is HTTP specific.

danielduong8:
Sorry for the last commit, should be ready for review now!

plausibly:
The routes seem to break when deploying to Heroku

danielduong8:
> The routes seem to break when deploying to Heroku

just need to add CI='false' env variable to heroku

server/routes/v1/tests/user.test.ts
```diff
@@ -0,0 +1,97 @@
...
+function mockResponse() {
+  const res: any = {};
+  res.status = jest.fn().mockReturnValue(res);
+  res.json = jest.fn().mockReturnValue(res);
+  res.cookie = jest.fn().mockReturnValue(res);
+  return res;
+}
+
+beforeAll(async () => {
+  await db.sequelize.sync({ force: true });
```
plausibly, line 20: there is a helper function that does this

server/routes/v1/user.ts
```diff
@@ -1,59 +1,129 @@
...
 
 const userRouter = express.Router();
-const eServ = new EmailService();
-const uContr = new UserController(db.User, eServ);
 
-userRouter.get(
-  "/confirmation/c=:token",
-  async (req: Request, res: Response) => {
-    const token = req.params.token;
+export const uContr: UserController = new UserController(
```
plausibly, line 12: shouldn't export this, it is only used in this file

server/services/emailService.ts
```diff
@@ -1,6 +1,9 @@
 import sgMail from "@sendgrid/mail";
 export default class EmailService {
-  constructor() {
+  apiRoute: string;
```
plausibly, line 3: private apiRoute

plausibly (approved):
EDIT: check changes requested

server/routes/v1/user.ts
```diff
@@ -1,59 +1,129 @@
...
+    user = response.data.result;
+
+    if (user) {
+      if (!process.env.SECRET) {
+        throw "Undefined secret";
+      }
+
+      const token = jwt.sign(
+        { username: user.userName, id: user.id },
+        process.env.SECRET,
```
plausibly, line 43: This should be called something else since cookie parser is also called SECRET

plausibly (changes requested):

server/services/emailService.ts
```diff
@@ -1,6 +1,9 @@
 import sgMail from "@sendgrid/mail";
 export default class EmailService {
-  constructor() {
+  private apiRoute: string;
+
+  constructor(apiRoute: string = `${process.env.PAGE_URL}api/v1/user`) {
```
andy9775, line 5:
```suggestion
  constructor(apiRoute: string ) {
```
This should be passed in to the function since there are two assumptions: 
1) The api version
2) the user route - what if other routes want to send emails
Best not to default these types of variables. Further, the `user` part of the route should not be passed in and it should be part of the built string in the function - just `/api/<version>`

server/server.ts
```diff
@@ -2,16 +2,16 @@ import express from "express";
...
-import userRouter from "./routes/v1/user";
+import v1Routes from "./routes/v1";
 
 const app = express();
 const port = 8080;
 
 app.use(express.json());
 app.use(express.urlencoded({ extended: true }));
-app.use(cookieParser(process.env.SECRET));
+app.use(cookieParser(process.env.COOKIE_SECRET));
```
andy9775, line 12:
```suggestion
app.use(cookieParser());
```
Since the JWT is signed, I don't think this is necessary. Also update README.md

server/routes/v1/user.ts
```diff
@@ -1,60 +1,140 @@
...
+);
 
-export async function signInHandler(
-  req: Request,
-  res: Response
-) {
-  const { userName, password } = req.body;
-  response = await userController.signIn(userName, password);
+let response;
+let user: User | undefined;
```
andy9775, line 18:
```suggestion
```
No globals

server/routes/v1/user.ts
```diff
@@ -1,60 +1,140 @@
...
 
-  if (response.data.token) {
-    res.cookie("token", JSON.stringify(response.data.token), {
+async function signOutHandler(req: Request, res: Response) {
```
andy9775, line 20: Shouldn't end in `Handler`. `signOut` is fine I think.

server/routes/v1/user.ts
```diff
@@ -1,60 +1,140 @@
...
+  res
+    .clearCookie("token", {
```
andy9775, line 22: make the cooke key a const.

server/routes/v1/user.ts
```diff
@@ -1,60 +1,140 @@
...
       httpOnly: true,
-      expires: moment().add(1, "hour").toDate(),
       signed: true,
-    });
-  }
+    })
+    .status(200)
+    .json({ message: "Signed out successfully" });
```
andy9775, line 27:
```suggestion
    .status(204);
```

server/routes/v1/user.ts
```diff
@@ -1,60 +1,140 @@
...
+}
 
-  res.status(response.status).json(response.data);
+export async function signInHandler(req: Request, res: Response) {
+  try {
+    const { userName, password } = req.body;
+
+    if (!userName || !password) {
+      res.status(400).json({ message: "Username or password is undefined" });
```
andy9775, line 35: You could use a ternary operator here and specify whether it is username, password, or both. Would make the message more helpful

server/routes/v1/user.ts
```diff
@@ -1,60 +1,140 @@
...
+      return;
+    }
+
+    response = await uContr.signIn(userName, password);
+    user = response.data.result;
+
+    if (user) {
+      if (!process.env.JWT_SECRET) {
```
andy9775, line 43: We should do this check upon the server booting up in server.ts, not here.

server/routes/v1/user.ts
```diff
@@ -1,60 +1,140 @@
...
-  if (response.data.token) {
-    res.cookie("token", JSON.stringify(response.data.token), {
-      httpOnly: true,
-      expires: moment().add(1, "hour").toDate(),
-      signed: true,
+  res.status(response.status).json(response.data);
+}
+
+async function confirmEmailHandler(req: Request, res: Response) {
+  const token = req.params.token;
```
andy9775, line 89: Isn't it just `c`?

server/routes/v1/user.ts
```diff
@@ -1,60 +1,140 @@
...
+
+  const status = await uContr.confirmEmail(token);
+
+  if (status) {
+    res.status(200).json({ message: "Email successfully confirmed" });
+  } else {
+    res
+      .status(400)
+      .json({ code: 400, message: "Token is invalid or expired." });
+  }
```
andy9775, line 104:
```suggestion

  if (await uContr.confirmEmail(token)) {
    res.status(204);
  } else {
    res
      .status(400)
      .json({ code: 400, message: "Token is invalid or expired." });
  }
```

server/routes/v1/user.ts
```diff
@@ -1,60 +1,140 @@
...
+    res.status(400).json({
+      code: 400,
+      message: "Missing token or password.",
     });
+    return;
   }
 
-  res.status(response.status).json(response.data);
+  const status = await uContr.resetPassword(
+    req.params.token,
```
andy9775, line 117: Isn't it just `r`?

server/routes/v1/user.ts
```diff
@@ -1,60 +1,140 @@
...
 if (process.env.CI && process.env.CI == "false") {
-  router.post("/signin", signInHandler);
-  router.post("/signup", signUpHandler);
+  userRouter.post("/signin", signInHandler);
+  userRouter.post("/signup", signUpHandler);
+  userRouter.post("/signout", signOutHandler);
+
+  userRouter.get("/confirmation/c=:token", confirmEmailHandler);
+  userRouter.get("/reset-password/r=:token", resetPassHandler);
 }
```
andy9775: Is the conditional necessary here?

server/routes/v1/index.ts
```diff
@@ -3,6 +3,6 @@ import userRouter from "./user";
 
 const routes = express.Router();
 
-routes.use("/users/", userRouter);
+routes.use("/user", userRouter);
```
andy9775, line 6:
```suggestion
routes.use("/users", userRouter);
```

server/routes/v1/tests/user.test.ts
```diff
@@ -54,31 +54,40 @@ describe("v1 - User Routes", () => {
...
   describe("/signin endpoint", () => {
-    test("If successful signup, returns 200 and creates cookie", async () => {
+    test("If successful signin, returns 200 and creates cookie", async () => {
       const data = {
         userName: "userName",
         password: "password",
       };
       const req = mockRequest(data) as Request;
       const res = mockResponse() as Response;
+      process.env.JWT_SECRET = "test";
```
andy9775, line 68: Why set this here?

andy9775 (changes requested):

server/controllers/v1/user.ts
```diff
@@ -16,6 +17,99 @@ export default class UserController {
...
+        email: email,
+        userName: userName,
+        password: hashedPassword,
+        firstName: firstName,
+        lastName: lastName,
+      });
+
+      await this.sendEmailConfirmation(result.email);
+
+      return { status: 201, data: { message: "Confirmation email sent" } };
```
andy9775, line 107: I think this can be a 204

server/controllers/v1/user.ts
```diff
@@ -16,6 +17,99 @@ export default class UserController {
...
+    } catch (error) {
+      console.error(error);
+      return { status: 500, data: { message: "Something went wrong" } };
```
andy9775, line 110: More descriptive message?

server/routes/v1/user.ts
```diff
@@ -1,59 +1,140 @@
...
+  if (!token) {
+    res.status(400).json({ code: 400, message: "Missing token." });
+    return;
   }
-);
+
+  const status = await uContr.confirmEmail(token);
+
+  if (status) {
+    res.status(200).json({ message: "Email successfully confirmed" });
```
andy9775, line 99: 204

server/routes/v1/user.ts
```diff
@@ -1,59 +1,140 @@
...
+  } else {
+    res
+      .status(400)
+      .json({ code: 400, message: "Token is invalid or expired." });
+  }
+}
+
+async function resetPassHandler(req: Request, res: Response) {
+  if (!req.params.token || !req.body.pass || !req.body.confpw) {
```
andy9775, line 108: Lets standardize on the naming convention. You use `password` in `signUpHandler` but `pass` here.

server/routes/v1/tests/user.test.ts
```diff
@@ -0,0 +1,97 @@
...
+        email: "email@mail.yahoo.ca",
+        userName: "userName",
+        password: "password",
+        firstName: "firstName",
+        lastName: "lastName",
+      };
+      const req = mockRequest(data) as Request;
+      const res = mockResponse() as Response;
+      await signUpHandler(req, res);
+      expect(res.status).not.toHaveBeenCalledWith(201);
```
andy9775, line 55: Can we assert a specific code? This is brittle since calling with 200, 204 will all pass

server/routes/v1/tests/user.test.ts
```diff
@@ -0,0 +1,97 @@
...
+    test("If unsuccessful signin, returns error code and does not create cookie", async () => {
+      const data = {
+        userName: "userName",
+        password: "wrongPassword",
+      };
+      const req = mockRequest(data) as Request;
+      const res = mockResponse() as Response;
+      process.env.JWT_SECRET = "test";
+      await signInHandler(req, res);
+      expect(res.status).not.toHaveBeenCalledWith(200);
```
andy9775, line 92: Same here.

andy9775 (changes requested):
Just a few more comments.

I've also noticed what we use `moment` in some places and `Date` in others - lets stay consistent. 

Also, you can getaway with using a 204. For things like signOut, it's implied that a 200 means that the signout was successful, so no need for the message. Then when the f/e gets the 204 it can decide what text to display to the user.

server/routes/v1/user.ts
danielduong8: pouya!!!

plausibly: :(

server/routes/v1/user.ts
```diff
@@ -1,60 +1,62 @@
...
     const { userName, password } = req.body;
 
     if (!userName || !password) {
-      res.status(400).json({ message: "Username or password is undefined" });
+      res.status(400).json({
+        message: !userName
+          ? !password
+            ? "Username and password"
+            : "Username"
+          : "Password" + " is undefined",
```
andy9775, line 35:
```suggestion
          : "Password" + " not provided",
```

This makes it easier for users to understand.

server/routes/v1/user.ts
```diff
@@ -105,7 +106,7 @@ async function confirmEmailHandler(req: Request, res: Response) {
 }
 
 async function resetPassHandler(req: Request, res: Response) {
-  if (!req.params.token || !req.body.pass || !req.body.confpw) {
+  if (!req.params.token || !req.body.password || !req.body.confpw) {
```
andy9775, line 109: `confpw` or should we use `password_confirmation`?

server/routes/v1/user.ts
andy9775: Forgot to address this.

server/routes/v1/user.ts
```diff
@@ -115,26 +116,24 @@ async function resetPassHandler(req: Request, res: Response) {
...
-  userRouter.post("/signout", signOutHandler);
+userRouter.post("/signin", signInHandler);
+userRouter.post("/signup", signUpHandler);
+userRouter.post("/signout", signOut);
 
-  userRouter.get("/confirmation/c=:token", confirmEmailHandler);
-  userRouter.get("/reset-password/r=:token", resetPassHandler);
-}
+userRouter.get("/confirmation/c=:token", confirmEmailHandler);
+userRouter.get("/reset-password/r=:token", resetPassHandler);
```
andy9775, line 137: You can't assign query params.

server/server.ts
```diff
@@ -4,17 +4,21 @@ import cookieParser from "cookie-parser";
 import db from "./models/index";
 import v1Routes from "./routes/v1";
 
-const app = express();
-const port = 8080;
+if (process.env.JWT_SECRET) {
```
andy9775, line 7:
```suggestion
if (!process.env.JWT_SECRET) {
    console.error(...);
    return 1;
}
```
It's better to have the bad path in an if block

server/middleware/auth.ts
```diff
@@ -1,23 +1,30 @@
...
+          where: {
+            id: decodedData?.id,
+          },
+        });
+        next();
+      } else {
+        res.status(401).json({ message: "Missing token in cookie" });
+      }
+    } catch (error) {
+      console.error(error);
```
andy9775, line 24: `jwt.verify` will throw an error on a bad token, but this doesn't mean that it's an internal server error. It should still be a 401 "Bad Session"(?)

server/middleware/auth.ts
```diff
@@ -1,23 +1,30 @@
...
+      const token = req.cookies.token;
 
-    let decodedData;
-
-    if (token) {
-      decodedData = jwt.verify(token, secret) as jwt.JwtPayload;
-      res.locals.id = decodedData?.id;
+      if (token) {
+        const decodedData = jwt.verify(token, secret) as jwt.JwtPayload;
+        res.locals.user = await userRepo.findOne({
```
andy9775, line 14: Should we add a helper function in this file to pull the user off? It would help for handling the case where the locals value is missing so we don't need the extra code in the handlers.

andy9775 (changes requested):
One additional comment.

You return the entire user model back to the api caller - this includes the hashed password. Most ORM's have a way to limit was is serialized - so lets ensure that the password isn't serialized, and lets write a simple test to ensure that the returned data doesn't include the password.

server/server.ts
danielduong8: I got an error when I tried to return here because it says I'm not inside a function block?

server/controllers/v1/user.ts
```diff
@@ -16,6 +17,109 @@ export default class UserController {
...
+        return { status: 403, data: { message: "Email has not been confirmed" }}
+      }
+
+      this.updateLastLogin(user);
+
+      user = await this.userRepo.findOne({
+        where: {
+          userName: userName,
+        },
+      });
```
andy9775, line 68: I don't think it's necessary to do a double query here. I would copy the original user object from the first query, delete the password field, and return that.

server/middleware/auth.ts
```diff
@@ -4,6 +4,15 @@ import { User } from "../models/user";
 
 const secret = process.env.JWT_SECRET as string;
 
+function getAuthUser(res: Response): User | undefined {
```
andy9775, line 7: Should be exported

server/middleware/auth.ts
```diff
@@ -4,6 +4,15 @@ import { User } from "../models/user";
...
+  if (!res.locals.user) {
+    res.status(401).json({ message: "User is not authorized to perform this request" })
+    return;
```
andy9775, line 10: Maybe this should raise an error? Create a new error class `Unauthorized`

danielduong8, line 10: throw error

andy9775 (approved):
Just a few comments, otherwise LGTM!

PR #21, opened 2021-10-29
Implement userRoute (password reset & confirmation)

plausibly:


server/routes/v1/index.ts
```diff
@@ -0,0 +1,8 @@
+import express from "express";
+import userRouter from "./user";
+
+const allRoutes = express.Router();
```
andy9775, line 4: nit:
```suggestion
const routes = express.Router();
```

server/routes/v1/user.ts
```diff
@@ -0,0 +1,42 @@
...
+  async (req: Request, res: Response) => {
+    const status: boolean = await uContr.confirmEmail(req.params.token);
+
+    if (status) {
+      res.status(200).send("Your email address has been confirmed. ");
+    }
+
+    res
+      .status(401)
+      .send("An error occurred. The link is either invalid or expired. ");
```
andy9775, line 21: you want to wrap this in an else block

server/server.ts
```diff
@@ -1,18 +1,21 @@
...
 const port = 8080;
- 
+
 app.use(express.json());
 app.use(express.urlencoded({ extended: true }));
 
 app.get("/v1/test", (req: express.Request, res: express.Response) => {
-    res.send({ express: 'Hello From Express' });
+  res.send({ express: "Hello From Express" });
 });
```
andy9775, line 13: we can remove this, I don't think we need the test route anymore

server/routes/v1/user.ts
```diff
@@ -0,0 +1,42 @@
...
+import UserController from "../../controllers/v1/user";
+
+const userRouter = express.Router();
+const eServ = new EmailService();
+const uContr = new UserController(db.User, eServ);
+
+userRouter.get(
+  "/confirmation/c=:token",
+  async (req: Request, res: Response) => {
+    const status: boolean = await uContr.confirmEmail(req.params.token);
```
andy9775, line 13: I would check if req.params.token is null. Return a 400. The error message (in the recommended response structure) would be something like: "Missing confirmation token."

server/routes/v1/user.ts
```diff
@@ -0,0 +1,42 @@
...
+
+    if (status) {
+      res.status(200).send("Your email address has been confirmed. ");
+    } else {
+      res
+        .status(401)
```
andy9775, line 19: A 401 indicates that the user needs to be authenticated in order to execute the request. But they can't be authenticated without confirming their account - should be 400. The message body should also show the error message, something like: "Invalid token. Please try again."

server/routes/v1/user.ts
```diff
@@ -0,0 +1,42 @@
...
+const eServ = new EmailService();
+const uContr = new UserController(db.User, eServ);
+
+userRouter.get(
+  "/confirmation/c=:token",
+  async (req: Request, res: Response) => {
+    const status: boolean = await uContr.confirmEmail(req.params.token);
+
+    if (status) {
+      res.status(200).send("Your email address has been confirmed. ");
```
andy9775, line 16:
```suggestion
      res.status(204);
```

Since it is our client that is handling the request, we can have the f/e show the message when it gets a 204 response.

server/routes/v1/user.ts
```diff
@@ -0,0 +1,42 @@
...
+    const status: boolean = await uContr.resetPassword(
+      req.params.token,
+      req.body.pass,
+      req.body.confpw
+    );
+
+    if (status) {
+      res.status(200).send("Your password has been changed!");
+    } else {
+      res.status(401).send("An error occurred.");
```
andy9775, line 37: Same comment as above for the confirmation route

server/routes/v1/user.ts
```diff
@@ -0,0 +1,42 @@
...
+  "/reset-password/r=:token",
+  async (req: Request, res: Response) => {
+    const status: boolean = await uContr.resetPassword(
+      req.params.token,
+      req.body.pass,
+      req.body.confpw
+    );
+
+    if (status) {
+      res.status(200).send("Your password has been changed!");
```
andy9775, line 35: Same comment as above for the confirmation route

server/routes/v1/user.ts
```diff
@@ -0,0 +1,42 @@
...
+  }
+);
+
+userRouter.get(
+  "/reset-password/r=:token",
+  async (req: Request, res: Response) => {
+    const status: boolean = await uContr.resetPassword(
+      req.params.token,
+      req.body.pass,
+      req.body.confpw
```
andy9775, line 31: check for missing/empty/null params.

server/routes/v1/user.ts
andy9775:
```suggestion
    const status = await uContr.confirmEmail(req.params.token);
```

I don't think we need to specify the type here.

server/server.ts
```diff
@@ -1,18 +1,17 @@
-import express from 'express';
-import db from './models/index';
+import express from "express";
+import db from "./models/index";
+import allRoutes from "./routes/v1";
```
andy9775, line 3:
```suggestion
import v1Routes from "./routes/v1";
```

andy9775 (changes requested):
Just some minor nits.

Since this is a JSON APi, you want to send out JSON responses.

You could have:
```JSON
{
code: <status_code>,
errors: <error message>,
data: <object>,
}
```

server/routes/v1/user.ts
plausibly: So for the success case we can do 204, and for the error case we can return a JSON object containing the error message and code?

danielduong8 (approved):

andy9775 (approved):
LGTM!

Issue #16, opened 2021-10-21
Setup the reverse proxy to the API. Currently we proxy to the react app, we need to proxy to the api subdomain. The api should also be versioned as part of the path.


plausibly:
for instance: api.example.com/v1/posts

danielduong8:
Since Heroku does not support subdomains, example.com/api/v1/posts is the format we are going with for now.

PR #22, opened 2021-10-31
Set up NGINX reverse proxy

danielduong8:
Set up reverse proxy to API #16 

- Update nginx config to proxy all URIs that start with /api/

### DIFF
--- a/nginx/conf.d/default.conf.template
+++ b/nginx/conf.d/default.conf.template
@@ -11,16 +11,9 @@ server {
     location = /50x.html {
         root   /usr/share/nginx/html;
     }
-}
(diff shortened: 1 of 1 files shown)


danielduong8:
Marking this as draft until #10 is approved.

plausibly (changes requested):
Deploy and test if the routes work

danielduong8:
> Deploy and test if the routes work

Waiting on #10 to be approved and merged since it has the /api/user/ endpoint and also the signup and signin logic.

andy9775 (approved):
LGTM!

PR #23, opened 2021-10-31
Fix broken deployment on node:17-alpine

danielduong8:
Use specific version 16.13-alpine instead of latest for Docker images

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,17 +1,17 @@
 # => Build client
-FROM node:alpine as client_builder
+FROM node:16-alpine as client_builder
 WORKDIR /app/client
 COPY client .
 RUN yarn install --production && yarn run build
 
 # => Build server
-FROM node:alpine as server_builder
+FROM node:16-alpine as server_builder
 WORKDIR /app/server
 COPY server .
 RUN yarn install && yarn run build
 
 # => Run container
-FROM nginx:alpine as base
+FROM nginx:1.20-alpine as base
 
 # Default port exposure
 EXPOSE 80


Dockerfile
```diff
@@ -1,17 +1,17 @@
 # => Build client
-FROM node:alpine as client_builder
+FROM node:16-alpine as client_builder
```
andy9775, line 2: Did `17` break things or `17.0.1`?

plausibly: When I tested this, node 17 was broken it gave some hash error. However, the other images weren't running due to a bash error:

Did you fix the `/bin/sh envsubst not found` issue? I was trying node images and was getting that error.

danielduong8: I think 17 and 17.0.1 are identical according to Docker Hub.

The following tags are also all equivalent: 17, 17-bullseye, 17.0, 17.0-bullseye, 17.0.1, 17.0.1-bullseye, bullseye, current, current-bullseye, latest.

plausibly: Deploy it on heroku and see if it works

danielduong8: Works

andy9775: ya, so 17 and latest all evaluate to 17.0.1. If 17.0.0 is fine, lets use that version.

danielduong8: I think 17.0.0 is also broken.

plausibly: The issue is with version 17 and above, so we need to use 16. https://github.com/webpack/webpack/issues/14532

plausibly (approved):

Dockerfile
andy9775: Sounds good!

andy9775 (approved):
lgtm

PR #26, opened 2021-11-02
Update `MaterialUI` and fix imports

plausibly:
- The @mui version we were using seemed to be outdated, this has been fixed along with the necessary imports
- Removed some unused imports
- Removed package-lock.json since we are using yarn

If you are working on mui with your branch, to update the import just change the material UI part to be `"@mui/material/"` instead of `@material-ui/core`


client/src/components/Login.tsx
```diff
@@ -1,7 +1,15 @@
 import React, { useState } from "react";
 
-import { Grid, Paper, Avatar, TextField, Button, Box } from "@material-ui/core";
-import LockOutlinedIcon from "@material-ui/icons/LockOutlined";
+
+import Grid from "@mui/material/Grid";
```
danielduong8, line 4: Not sure why you split the imports up like this, I tested it and they should still work fine if you do:

`import { ... } from "@mui/material

client/src/components/SignUp.tsx
```diff
@@ -1,8 +1,12 @@
...
-import CircleOutlined from "@material-ui/icons/AddCircleOutlineOutlined";
-import useMediaQuery from "@material-ui/core/useMediaQuery";
+import Grid from "@mui/material/Grid";
+import Paper from "@mui/material/Paper";
+import Avatar from "@mui/material/Avatar";
+import TextField from "@mui/material/TextField";
+import Button from "@mui/material/Button";
+import Box from "@mui/material/Box";
+import CircleOutlined from "@mui/icons-material/AddCircleOutlineOutlined";
+import useMediaQuery from "@mui/material/useMediaQuery";
```
danielduong8, line 9: Same comment here as above.

client/src/containers/AuthContainer.tsx
```diff
@@ -1,13 +1,14 @@
 import React, { useState } from "react";
+ 
+import Tabs from "@mui/material/Tabs";
+import Tab from "@mui/material/Tab";
 
-import Tabs from "@material-ui/core/Tabs";
-import Tab from "@material-ui/core/Tab";
+import Box from "@mui/material/Box";
+import Paper from "@mui/material/Paper";
+import Grid from "@mui/material/Grid";
```
danielduong8, line 8: Here as well

danielduong8 (changes requested):

client/src/components/Login.tsx
plausibly: The build time becomes insanely slow when I do that

client/src/components/SignUp.tsx
plausibly: no.

client/src/containers/AuthContainer.tsx
plausibly: no!

client/src/components/Login.tsx
danielduong8: I looked it up and I think it has something to do with "tree-shaking".

danielduong8: > What we've learned is that in order to take advantage of tree shaking, you must...
> 
> Use ES2015 module syntax (i.e. import and export).
> Ensure no compilers transform your ES2015 module syntax into CommonJS modules (this is the default behavior of the popular Babel preset @babel/preset-env - see the documentation for more details).
> Add a "sideEffects" property to your project's package.json file.
> Use the production mode configuration option to enable various optimizations including minification and tree shaking.

We could do this but it's probably just easier to separate imports.

client/src/App.test.tsx
```diff
@@ -1,4 +1,3 @@
-import React from "react";
 import { render, screen } from "@testing-library/react";
 import App from "./App";
```
andy9775, line 3: Why is removing the react import necessary? If it's not, lets leave it out of this PR.

client/src/App.tsx
```diff
@@ -1,4 +1,3 @@
-import React from "react";
 import "./App.css";
 
 import { AuthContainer } from "./containers";
```
andy9775, line 3: ditto

client/src/components/Login.tsx
andy9775: "Tree shaking" removes unused imports and makes your bundle size smaller (the JS file that's built) which makes for quicker asset download times and a better user experience overall.

client/src/containers/AuthContainer.test.tsx
```diff
@@ -1,8 +1,6 @@
-import React from "react";
-
```
andy9775, line 2: necessary?

client/package.json
```diff
@@ -3,8 +3,11 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
-    "@material-ui/core": "^4.12.3",
-    "@material-ui/icons": "^4.11.2",
+    "@babel/core": "^7.16.0",
+    "@emotion/react": "^11.5.0",
+    "@emotion/styled": "^11.3.0",
+    "@mui/icons-material": "^5.0.5",
```
andy9775, line 9: Can we skip the icons if we're not using them

client/package.json
```diff
@@ -3,8 +3,11 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
-    "@material-ui/core": "^4.12.3",
-    "@material-ui/icons": "^4.11.2",
+    "@babel/core": "^7.16.0",
+    "@emotion/react": "^11.5.0",
+    "@emotion/styled": "^11.3.0",
```
andy9775, line 8: Does mui use styled components? This is a whole other library

andy9775 (changes requested):
Just a few minor comments.

client/package.json
plausibly: It is required and was yarn was getting mad without them

client/package.json
plausibly: Yes, they give the option of installing this or `styled-components`

client/src/App.test.tsx
```diff
@@ -1,6 +1,6 @@
...
+import React from "react";
```
andy9775, line 3:
```suggestion
import React from "react";
import { render, screen } from "@testing-library/react";
import App from "./App";
```

client/src/App.tsx
```diff
@@ -1,6 +1,6 @@
-import React from "react";
 import "./App.css";
 
+import React from "react";
```
andy9775, line 3:
```suggestion
import React from "react";
import "./App.css";
```

client/src/containers/AuthContainer.test.tsx
andy9775: Missed one

andy9775 (approved):
Just a few more comments, otherwise LGTM!

PR #28, opened 2021-11-04
Upgrade login page design, fix resizing issues

danielduong8:
- Use theming with createTheme instead of hardcoding colour values

- Use CssBaseline to provide normalized CSS stylings

- Fix ts error caused by conditional expect in App.test.tsx

- Use a hook to dynamically update signin and signup box sizes
  - Updates based on the size of the current viewport, so boxes can
adjust in realtime on screen size change.

- Allow scrolling to resolve issue of overflowing elements

- Add "Remember me" checkbox
  - This should be implemented in the backend in a future PR.

- Add "Forgot password?" link for password reset
  - This should be connected to the backend in a future PR.

- Update style of signup redirect and signin redirect buttons
  - Instead of using a Button style, use a Link component style.

- Update size of TextFields to be smaller
  - This will help with sizing errors and make things look more
compact.

- Remove unnecessary TextField placeholders
  - Only email placeholder is really necessary to give an example of
a valid email address. It can be assumed that most people know how
to input their name, username, and password.

- Update background image and its css properties
  - Image needs to stretch to the full viewport.

- Use media query with breakpoint to hide "A place to connect"
  - This text does not fit and causes overflow when the screen size
is too small. Only make the text visible to those who have a
sufficiently large viewport or screen.

- Set overflow to hidden for Grid item containing Tabs
  - This should prevent the ripple effect, when clicking a tab, from
going past the rounded edges.

- Fix typos and sentence structure

image
image


client/src/containers/AuthContainer.tsx
```diff
@@ -29,70 +31,70 @@ function TabPanel(props: TabPanelProps) {
...
+            sx={{
               textAlign: "center",
-              paddingTop: matches ? "100px" : "0px",
             }}
-            fontFamily="Verdana"
-            color="white"
           >
-            <h1 style={{ fontSize: matches ? "500%" : "200%" }}>UOFTBOARD</h1>
-            <h1>A place to connect</h1>
+            <Typography color="grey" fontFamily="Verdana" variant="h2">
```
plausibly, line 65: Change the font color, it is clashing with the background.

plausibly (changes requested):

client/src/containers/AuthContainer.tsx
danielduong8: I am aware but looking through all the available colours is hard, unless you know of a good colour to use

danielduong8: Fixed

client/src/App.test.tsx
```diff
@@ -4,6 +4,6 @@ import App from "./App";
 
 test("renders learn react link", () => {
   render(<App />);
-  const input = screen.getByText("Log In");
+  const input = screen.getByText("Sign in");
```
andy9775, line 7: I think this is already a test, no?

client/src/App.tsx
```diff
@@ -1,12 +1,25 @@
 import React from "react";
 import "./App.css";
 import { AuthContainer } from "./containers";
+import CssBaseline from "@mui/material/CssBaseline";
+import { createTheme, ThemeProvider } from "@mui/material/styles";
+
+const theme = createTheme({
+  palette: {
+    secondary: {
+      main: "#f4003d",
```
andy9775, line 10: Dark mode as a stretch goal?

client/src/components/Login.tsx
```diff
@@ -39,80 +32,97 @@ function Login(props: { handleChange: Function }) {
...
+    <Box
+      data-testid="LogInTab"
+      sx={{
+        mx: 2,
+        display: "flex",
+        flexDirection: "column",
+        alignItems: "center",
+      }}
+      overflow="auto"
+      maxHeight={height - 200 > 0 ? height - 200 : 0}
```
andy9775, line 58: It's bad practice to calculate the height since it happens on each render. Can we make it full height, then have a padding, or margin?

client/src/components/SignUp.tsx
```diff
@@ -81,27 +71,36 @@ function SignUp(props: { handleChange: Function }) {
...
+    <Box
+      data-testid="SignUpTab"
+      sx={{
+        mx: 2,
+        display: "flex",
+        flexDirection: "column",
+        alignItems: "center",
+      }}
+      overflow="auto"
+      maxHeight={height - 200 > 0 ? height - 200 : 0}
```
andy9775, line 83: Ditto

client/src/containers/AuthContainer.css
```diff
@@ -1,5 +1,8 @@
 .AuthContainer {
-  background-color: #488fe0;
+  background: url("https://cdn.skyrisecities.com/sites/default/files/images/projects/6977/6977-78314.jpeg");
```
andy9775, line 2: Do we have a licence?

andy9775 (changes requested):
Just a few comments after a first pass.

client/src/App.tsx
danielduong8: Will add this to the projects board as a goal, but I think for now we should probably focus on just getting the UI correctly set up first and functioning.

client/src/containers/AuthContainer.css
danielduong8: Couldn't find any good pictures on unsplash, is it fine if we leave this as a background for now since we're still in development and aren't going to be releasing it anytime soon?

plausibly: The image was taken by University of Toronto, our project is for University of Toronto.. I think we may be allowed to use it

client/src/components/Login.tsx
danielduong8: I believe the height is only calculated in the event that the window is resized. 

I have a hook in window.tsx that only calculates the height when an resize window eventListener is triggered.

client/src/App.tsx
```diff
@@ -1,25 +1,23 @@
...
+      <Routes>
+        <Route path="/" element={<AuthContainer />} />
+        <Route
+          path="*"
+          element={
+            <main style={{ padding: "1rem" }}>
+              <p>There's nothing here!</p>
+            </main>
+          }
+        />
```
andy9775, line 18: Not sure if this is necessary. Having a `404 - not found` page is good, but it should be cleaner.

client/src/containers/AuthContainer.tsx
```diff
@@ -49,63 +57,66 @@ function AuthContainer() {
...
-                UBoard
-              </Typography>
-              <Typography color="white" fontFamily="Verdana" variant="h4">
-                A place to connect
-              </Typography>
-            </Box>
-            
-          </Box>
-        </Grid>
+      <ThemeProvider theme={theme}>
```
andy9775, line 60: The theme provider should be at the root as the theme is applied application wide. So it should wrap all routes - not just this one.

client/src/containers/AuthContainer.css
andy9775: I don't think we are. Despite the project being a UofT project, we would still need permission to use their assets. This could raise some questions during demo at the very least.

client/src/containers/AuthContainer.tsx
```diff
@@ -29,72 +41,82 @@ function TabPanel(props: TabPanelProps) {
...
-
-            <TabPanel value={tabIndex} index={0}>
-              <Login handleChange={handleChange} />
-            </TabPanel>
-            <TabPanel value={tabIndex} index={1}>
-              <SignUp handleChange={handleChange} />
-            </TabPanel>
-          </Paper>
+              <Box bgcolor="rgba(83, 102, 133, 0.6)" p="40px" px="80px">
+                <Typography color="white" fontFamily="Verdana" variant="h2">
```
andy9775, line 79: typography would be a good theme variable.

andy9775 (changes requested):
Just a few more comments.

client/src/App.tsx
danielduong8: Right, this is probably something that can be added to nginx later instead of here.

client/src/containers/AuthContainer.tsx
danielduong8: Ok, I think should Pouya might have to update his PR as well then, since it looks like he might be overriding the main theme here.

client/src/containers/AuthContainer.tsx
danielduong8: Decided not to use Verdana, so theme for Typography is currently undecided.

plausibly: I think we can stick with the theme default, we don't need to declare fontFamily in any of our typography calls.

client/src/containers/AuthContainer.tsx
plausibly: Ok

danielduong8:
The new bg image is from Wikimedia Commons so we should be good I think.

client/src/App.tsx
```diff
@@ -2,23 +2,30 @@ import React from "react";
 import { BrowserRouter, Route, Routes } from "react-router-dom";
 import "./App.css";
 import { AuthContainer } from "./containers";
+import PostDashboard from "./containers/PostDashboard";
+
+import CssBaseline from "@mui/material/CssBaseline";
+import { createTheme, ThemeProvider } from "@mui/material/styles";
```
andy9775: Dependency imports should go above local imports.

andy9775 (approved):
Are we not displaying the title when viewed on smaller screens?
Otherwise LGTM!

danielduong8:
I'm not sure, I think we still might create an actual landing page in the future, but otherwise, I might make a separate PR to implement that because I want to implement some sort of sliding window thing.

For now, though I think I want to get this merged so that I can connect the backend and have stuff good to go.

PR #31, opened 2021-11-05
Creating a new post and implemented a popup preview for the post

gurvirsasan:
- added `CreatePost.tsx` which is a form component that asks user for post information such as title, body, image, tags, location, and capacity
- added a preview feature for the post in `PreviewPopUp.tsx`

### DIFF
--- a/client/src/components/CreatePost.test.tsx
+++ b/client/src/components/CreatePost.test.tsx
@@ -2,7 +2,7 @@ import React from "react";
 import CreatePost from "./CreatePost";
 import { unmountComponentAtNode } from "react-dom";
 import { act } from "react-dom/test-utils";
-import { render, screen, fireEvent } from "@testing-library/react";
+import { render, screen, fireEvent, cleanup } from "@testing-library/react";
 import newImage from "../assets/background.jpg";
 
 let container: HTMLElement | null = null;
@@ -10,6 +10,9 @@ beforeEach(() => {
   // setup a DOM element as a render target
   container = document.createElement("div");
   document.body.appendChild(container);
+  act(() => {
+    render(<CreatePost />);
+  });
 });
 
 afterEach(() => {
@@ -18,26 +21,17 @@ afterEach(() => {
     unmountComponentAtNode(container);
     container.remove();
     container = null;
+    cleanup();
   }
 });
 
 describe("verifying launch of create post component", () => {
-  it("should render out the create post form", () => {
-    act(() => {
-      render(<CreatePost />);
-    });
-    const bodyTextField = screen.getByTestId("bodyTextField");
-    expect(bodyTextField).toBeInTheDocument();
+  it("shows the preview button as disabled", () => {
+    // get the preview button
+    expect(screen.getByTestId("previewButton")).toBeDisabled();
   });
 
-  it("shows the preview button as disabled", () => {
-    act(() => {
-      render(<CreatePost />);
-    });
-
-    // get the preview button
-    const previewButton = screen.getByTestId("previewButton");
-    expect(previewButton).toBeDisabled();
+  it("should undisable the preview button after providing required fields", () => {
     // input title
     const titleTextField = screen.getByPlaceholderText("title");
     fireEvent.change(titleTextField, { target: { value: "Test Club" } });
@@ -45,6 +39,6 @@ describe("verifying launch of create post component", () => {
     const bodyTextField = screen.getByPlaceholderText("description");
     fireEvent.change(bodyTextField, { target: { value: "body" } });
     fireEvent.blur(bodyTextField);
-    expect(previewButton).not.toBeDisabled();
+    expect(screen.getByTestId("previewButton")).not.toBeDisabled();
   });
 });
--- a/client/src/components/CreatePost.test.tsx
+++ b/client/src/components/CreatePost.test.tsx
@@ -31,7 +31,7 @@ describe("verifying launch of create post component", () => {
     expect(screen.getByTestId("previewButton")).toBeDisabled();
   });
 
-  it("should undisable the preview button after providing required fields", () => {
+  it("should enable the preview button after providing required fields", () => {
     // input title
     const titleTextField = screen.getByPlaceholderText("title");
     fireEvent.change(titleTextField, { target: { value: "Test Club" } });
@@ -42,7 +42,7 @@ describe("verifying launch of create post component", () => {
     expect(screen.getByTestId("previewButton")).not.toBeDisabled();
   });
 
-  it("renders `previewPopUp component upon clicking Preview", () => {
+  it("renders `previewPopUp` component upon clicking Preview", () => {
     // input title
     fireEvent.change(screen.getByPlaceholderText("title"), {
       target: { value: "Test Club" },
--- a/client/src/components/CreatePost.test.tsx
+++ b/client/src/components/CreatePost.test.tsx
@@ -41,4 +41,18 @@ describe("verifying launch of create post component", () => {
     fireEvent.blur(bodyTextField);
     expect(screen.getByTestId("previewButton")).not.toBeDisabled();
   });
+
+  it("renders `previewPopUp component upon clicking Preview", () => {
+    // input title
+    fireEvent.change(screen.getByPlaceholderText("title"), {
+      target: { value: "Test Club" },
+    });
+    // input body
+    const bodyTextField = screen.getByPlaceholderText("description");
+    fireEvent.change(bodyTextField, { target: { value: "body" } });
+    fireEvent.blur(bodyTextField);
+    // pop up should render
+    screen.getByTestId("previewButton").click();
+    expect(screen.getByTestId("PreviewPopUpComponent")).toBeInTheDocument();
+  });
 });
--- a/client/src/components/CreatePost.test.tsx
+++ b/client/src/components/CreatePost.test.tsx
@@ -1,0 +1,50 @@
+import React from "react";
+import CreatePost from "./CreatePost";
+import { unmountComponentAtNode } from "react-dom";
+import { act } from "react-dom/test-utils";
+import { render, screen, fireEvent } from "@testing-library/react";
+import newImage from "../assets/background.jpg";
+
+let container: HTMLElement | null = null;
+beforeEach(() => {
+  // setup a DOM element as a render target
+  container = document.createElement("div");
+  document.body.appendChild(container);
+});
+
+afterEach(() => {
+  // cleanup on exiting
+  if (container != null) {
+    unmountComponentAtNode(container);
+    container.remove();
+    container = null;
+  }
+});
+
+describe("verifying launch of create post component", () => {
+  it("should render out the create post form", () => {
+    act(() => {
+      render(<CreatePost />);
+    });
+    const bodyTextField = screen.getByTestId("bodyTextField");
+    expect(bodyTextField).toBeInTheDocument();
+  });
+
+  it("shows the preview button as disabled", () => {
+    act(() => {
+      render(<CreatePost />);
+    });
+
+    // get the preview button
+    const previewButton = screen.getByTestId("previewButton");
+    expect(previewButton).toBeDisabled();
+    // input title
+    const titleTextField = screen.getByPlaceholderText("title");
+    fireEvent.change(titleTextField, { target: { value: "Test Club" } });
+    // input body
+    const bodyTextField = screen.getByPlaceholderText("description");
+    fireEvent.change(bodyTextField, { target: { value: "body" } });
+    fireEvent.blur(bodyTextField);
+    expect(previewButton).not.toBeDisabled();
+  });
+});
--- a/client/src/components/CreatePost.tsx
+++ b/client/src/components/CreatePost.tsx
@@ -9,13 +9,9 @@ import Button from "@mui/material/Button";
 import TextField from "@mui/material/TextField";
 import Divider from "@mui/material/Divider";
 import Typography from "@mui/material/Typography";
-
-import Dialog from "@mui/material/Dialog";
-import DialogContent from "@mui/material/DialogContent";
-import DialogActions from "@mui/material/DialogActions";
+import Tooltip from "@mui/material/Tooltip";
 
 import PreviewPopUp from "./PreviewPopUp";
-import { DialogTitle } from "@mui/material";
 
 const theme = createTheme();
 
@@ -29,24 +25,12 @@ function CreatePost() {
   const [eventLocation, setEventLocation] = useState("");
 
   const [openPopup, setOpenPopup] = useState(false);
-  const [openPopupError, setOpenPopupError] = useState(false);
 
   const allFields = [title, body, file];
 
   // handle functions
   const handleClickOpen = () => {
-    let noEmptyFields = true;
-    allFields.forEach((inputField) => {
-      if (inputField === "") {
-        noEmptyFields = false;
-      }
-    });
-    if (noEmptyFields) {
-      setOpenPopup(true);
-    } else {
-      setOpenPopupError(true);
-    }
-    return;
+    setOpenPopup(true);
   };
 
   const handleClose = () => {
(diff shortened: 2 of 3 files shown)


plausibly (commented):
LGTM

gurvirsasan:
Create Post interface:
Screen Shot 2021-11-04 at 11 32 18 PM

Preview:
Screen Shot 2021-11-04 at 11 32 09 PM

Error Popup:
Screen Shot 2021-11-04 at 11 32 24 PM

client/src/components/CreatePost.tsx
```diff
@@ -0,0 +1,225 @@
...
+            tags={tags}
+            eventCapacity={eventCapacity}
+            location={eventLocation}
+            openPopup={openPopup}
+            handleClose={handleClose}
+          ></PreviewPopUp>
+
+          {/* Display Error */}
+
+          <Dialog open={openPopupError}>
```
plausibly, line 191: Nit - instead of an error dialog it might be better to use an alert or snackbar 

https://mui.com/components/alert/
https://mui.com/components/snackbars/

client/src/components/CreatePost.tsx
```diff
@@ -0,0 +1,225 @@
...
+function CreatePost() {
+  // create hooks for input fields
+  const [title, setTitle] = useState("");
+  const [body, setBody] = useState("");
+  const [file, setFile] = useState("");
+  const [tags, setTags] = useState("");
+  const [eventCapacity, setEventCapacity] = useState("");
+  const [eventLocation, setEventLocation] = useState("");
+
+  const [openPopup, setOpenPopup] = useState(false);
```
plausibly, line 31: If the actual popup is in `PreviewPopUp` then we should not be defining this here, the same idea applies to handleClose

client/src/components/CreatePost.tsx
```diff
@@ -0,0 +1,225 @@
...
+  const [openPopupError, setOpenPopupError] = useState(false);
```
plausibly, line 32: We can use an error alert/snackbar instead of an error popup, or just do nothing and highlight that fields are missing

plausibly (changes requested):
Few comments, otherwise it looks good

client/src/components/CreatePost.tsx
gurvirsasan: PreviewPopUp is just the component for the pop up, in order to trigger it, we require the hooks that need to be present in `CreatePost.tsx` because that's where the actual "preview" button is.

plausibly: Fair enough. Request a re-review from me once you go over the other comments

client/src/components/CreatePost.tsx
```diff
@@ -0,0 +1,225 @@
...
+import TextField from "@mui/material/TextField";
+import Divider from "@mui/material/Divider";
+import Typography from "@mui/material/Typography";
+
+import Dialog from "@mui/material/Dialog";
+import DialogContent from "@mui/material/DialogContent";
+import DialogActions from "@mui/material/DialogActions";
+
+import PreviewPopUp from "./PreviewPopUp";
+import { DialogTitle } from "@mui/material";
```
andy9775, line 18:
```suggestion

import { DialogTitle } from "@mui/material";

import PreviewPopUp from "./PreviewPopUp";
```

client/src/components/CreatePost.tsx
```diff
@@ -0,0 +1,225 @@
...
+const theme = createTheme();
+
+function CreatePost() {
+  // create hooks for input fields
+  const [title, setTitle] = useState("");
+  const [body, setBody] = useState("");
+  const [file, setFile] = useState("");
+  const [tags, setTags] = useState("");
+  const [eventCapacity, setEventCapacity] = useState("");
+  const [eventLocation, setEventLocation] = useState("");
```
andy9775, line 29: I believe @danielduong8 is storing form state in an object, lets stay consistent across pages. I'm fine with either approach.

client/src/components/CreatePost.tsx
```diff
@@ -0,0 +1,225 @@
...
+      if (inputField === "") {
+        noEmptyFields = false;
+      }
+    });
+    if (noEmptyFields) {
+      setOpenPopup(true);
+    } else {
+      setOpenPopupError(true);
+    }
+    return;
```
andy9775, line 49:
```suggestion
    if (allFields.map(e => !e.trim()).some(e => e)) { // if any input field is empty
          setOpenPopupError(true);
    } else {
      setOpenPopup(true);
    }
```

`allFields.map(e => !e.trim()).some(e => e)` will return `true` if one or more strings are empty. Has [good support](https://caniuse.com/mdn-javascript_builtins_array_some) as well

client/src/components/CreatePost.tsx
```diff
@@ -0,0 +1,225 @@
...
+  };
+
+  const handleClose = () => {
+    setOpenPopup(false);
+  };
```
andy9775, line 54: Since this is only used in one spot, I think we can make it an anonymous function at the call site.

client/src/components/CreatePost.tsx
```diff
@@ -0,0 +1,225 @@
...
+
+  // handle functions
+  const handleImageUpload = (event: React.ChangeEvent<{}>) => {
+    const target = event.target as HTMLInputElement;
+    /** do something with the file **/
+    let url = URL.createObjectURL((target.files as FileList)[0]);
+    setFile(url);
+    console.log(url);
+  };
```
andy9775, line 63: We can split the file up, then have the backend handle the file upload in parallel - should speed up file uploads significantly. 

reference: https://api.video/blog/tutorials/uploading-large-files-with-javascript

client/src/components/CreatePost.tsx
```diff
@@ -0,0 +1,225 @@
...
+                  size="small"
+                  onChange={(e) => setEventCapacity(e.target.value)}
+                />
+              </Grid>
+
+              <Grid item xs={12} md={5}>
+                <TextField
+                  fullWidth
+                  label="Location"
+                  placeholder="Zoom: https://utoronto.zoom.us/j/7227206933#success"
```
andy9775, line 130: I would advise against this type of a placeholder. Perhaps just "Deerfield Hall" or something?

client/src/components/PreviewPopUp.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+import Card from "@mui/material/Card";
+import CardContent from "@mui/material/CardContent";
+import CardMedia from "@mui/material/CardMedia";
+
+import Dialog from "@mui/material/Dialog";
+import DialogContent from "@mui/material/DialogContent";
+import DialogActions from "@mui/material/DialogActions";
+
+export default function PreviewPopUp(props: any) {
```
andy9775, line 15: Don't use `any`

client/src/components/CreatePost.tsx
```diff
@@ -0,0 +1,225 @@
...
+                </Button>
+              </Grid>
+              <Grid item xs={7} md={5}>
+                <Button
+                  fullWidth
+                  variant="contained"
+                  sx={{ mt: 3, mb: 2 }}
+                  size="large"
+                >
+                  Post
```
andy9775, line 172:
```suggestion
                  Create
```

andy9775 (changes requested):
Instead of having a pop-up telling the user that they have missing fields, do we want to grey out the "post" button and on-hover have a tool tip telling the user that there are missing fields? Seems more intuitive than having to deal with additional modals.

Also, no tests???

client/src/components/CreatePost.tsx
```diff
@@ -0,0 +1,183 @@
...
+    setOpenPopup(true);
+  };
+
+  // handle functions
+  const handleImageUpload = (event: React.ChangeEvent<{}>) => {
+    const target = event.target as HTMLInputElement;
+    /** do something with the file **/
+    let url = URL.createObjectURL((target.files as FileList)[0]);
+    setFile(url);
+    console.log(url);
```
andy9775, line 40:
```suggestion

```

andy9775 (changes requested):
There's a couple of things I think you missed. If you've addressed them, resolve the comment, or if you disagree explain why.

Also, can we get f/e routing setup before this? I'd like to play around with it locally before approving. It should be easy to get in.

client/src/components/CreatePost.tsx
gurvirsasan: Can you explain what you mean by storing form state in an object? I'm not familiar

client/src/components/CreatePost.test.tsx
```diff
@@ -0,0 +1,50 @@
...
+  if (container != null) {
+    unmountComponentAtNode(container);
+    container.remove();
+    container = null;
+  }
+});
+
+describe("verifying launch of create post component", () => {
+  it("should render out the create post form", () => {
+    act(() => {
```
plausibly, line 26: Instead of repeating this for each test we can put it into beforeEach, and in afterEach you can call cleanup() to destroy it

client/src/components/CreatePost.test.tsx
```diff
@@ -0,0 +1,50 @@
...
+      render(<CreatePost />);
+    });
+    const bodyTextField = screen.getByTestId("bodyTextField");
```
plausibly, line 29: Variable is unnecessary here

client/src/components/CreatePost.test.tsx
```diff
@@ -0,0 +1,50 @@
...
+    expect(bodyTextField).toBeInTheDocument();
+  });
+
+  it("shows the preview button as disabled", () => {
+    act(() => {
```
plausibly, line 34: put in BeforeEach

plausibly (changes requested):

client/src/components/CreatePost.tsx
danielduong8: e.g.
```
const [form, setForm] = useState({
    title: "",
    body: "",
    file: "",
    tags: "",
    eventCapacity: "",
    ....
  });
```

danielduong8: To set things you would have to use something like:
```
setForm({ ...form, [e.target.name]: e.target.value });
```
Keep in mind you would also have to set the `name=blah` properties on the components where you use this handler.

danielduong8: If you're not comfortable with doing it, you can leave this for now and I can take care of it once I issue the PR for connecting all the backend to the frontend.

gurvirsasan: okay that would be good, but I think I get it though

client/src/components/CreatePost.test.tsx
```diff
@@ -0,0 +1,50 @@
...
+  // cleanup on exiting
+  if (container != null) {
+    unmountComponentAtNode(container);
+    container.remove();
+    container = null;
+  }
+});
+
+describe("verifying launch of create post component", () => {
+  it("should render out the create post form", () => {
```
andy9775, line 25: Not really a necessary test. You should be testing things happening.

client/src/components/CreatePost.test.tsx
```diff
@@ -0,0 +1,50 @@
...
+    act(() => {
+      render(<CreatePost />);
+    });
+    const bodyTextField = screen.getByTestId("bodyTextField");
+    expect(bodyTextField).toBeInTheDocument();
+  });
+
+  it("shows the preview button as disabled", () => {
```
andy9775, line 33: Should also add a test for the preview button being enabled after the fields are filled out in a separate test.

andy9775 (changes requested):
Just a few points of feedback. Are there anymore test cases you think we may need?

client/src/components/CreatePost.test.tsx
```diff
@@ -18,33 +21,38 @@ afterEach(() => {
...
-      render(<CreatePost />);
-    });
-
     // get the preview button
-    const previewButton = screen.getByTestId("previewButton");
-    expect(previewButton).toBeDisabled();
+    expect(screen.getByTestId("previewButton")).toBeDisabled();
+  });
+
+  it("should undisable the preview button after providing required fields", () => {
```
andy9775, line 34:
```suggestion
  it("should enable the preview button after providing required fields", () => {
```

client/src/components/CreatePost.test.tsx
```diff
@@ -18,33 +21,38 @@ afterEach(() => {
...
     fireEvent.change(titleTextField, { target: { value: "Test Club" } });
     // input body
     const bodyTextField = screen.getByPlaceholderText("description");
     fireEvent.change(bodyTextField, { target: { value: "body" } });
     fireEvent.blur(bodyTextField);
-    expect(previewButton).not.toBeDisabled();
+    expect(screen.getByTestId("previewButton")).not.toBeDisabled();
+  });
+
+  it("renders `previewPopUp component upon clicking Preview", () => {
```
andy9775, line 45:
```suggestion
  it("renders `previewPopUp` component upon clicking Preview", () => {
```

client/src/components/CreatePost.tsx
andy9775: Still TODO?

client/src/components/PreviewPopUp.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+export default function PreviewPopUp(props: {
+  title: string;
+  body: string;
+  img: string;
+  tags: string;
+  eventCapacity: string;
+  location: string;
+  openPopup: boolean;
+  handleClose: any;
```
andy9775, line 23: This should be a function type. For example, `(a: string) => void` would be the annotation for a function which takes a string (one param) and returns nothing.

andy9775 (changes requested):

client/src/components/CreatePost.tsx
```diff
@@ -0,0 +1,188 @@
...
+              <Grid item xs={12}>
+                <TextField
+                  required
+                  fullWidth
+                  label="Body"
+                  data-testid="bodyTextField"
+                  placeholder="description"
+                  multiline
+                  rows={6}
+                  onChange={(e) => setForm({ ...form, body: e.target.value })}
```
danielduong8, line 83: Yeah this should work as well I believe no problem

client/src/components/CreatePost.tsx
```diff
@@ -36,147 +36,148 @@ function CreatePost() {
...
-                <TextField
-                  fullWidth
-                  label="Tags (Separate tags by a comma)"
-                  placeholder="Clubs, Math, MCS"
-                  size="small"
-                  onChange={(e) => setTags(e.target.value)}
-                />
-              </Grid>
+    <Container component="main" maxWidth="md">
+      <CssBaseline />
```
andy9775, line 44: Do we need this here?

client/src/components/CreatePost.tsx
```diff
@@ -36,147 +36,148 @@ function CreatePost() {
...
+                fullWidth
+                label="Tags (Separate tags by a comma)"
+                placeholder="Clubs, Math, MCS"
+                size="small"
+                onChange={(e) => setForm({ ...form, tags: e.target.value })}
+              />
+            </Grid>
+          </Grid>
+
+          {/* form  done */}
```
andy9775, line 132: Is this comment in the correct place?

client/src/components/CreatePost.tsx
```diff
@@ -36,147 +36,148 @@ function CreatePost() {
...
+          <Grid container spacing={2} justifyContent="center">
+            <Grid item xs={5} md={3}>
+              <Tooltip title="Enter all required fields!">
+                <Box>
+                  <Button
+                    fullWidth
+                    variant="contained"
+                    sx={{ mt: 3, mb: 2 }}
+                    color="secondary"
+                    onClick={() => handleClickOpen()}
```
andy9775, line 146:
```suggestion
                    onClick={handleClickOpen}
```

client/src/components/PreviewPopUp.tsx
```diff
@@ -20,7 +20,7 @@ export default function PreviewPopUp(props: {
   eventCapacity: string;
   location: string;
   openPopup: boolean;
-  handleClose: any;
+  handleClose: () => any;
```
andy9775, line 23:
```suggestion
  handleClose: () => void;
```

Does the function actually return `any`?

andy9775 (approved):
Just some minor comments. Please address, otherwise LGTM!

client/src/components/CreatePost.tsx
gurvirsasan: by the comment, I meant the input fields or form were done, the next components were just the buttons

client/src/components/PreviewPopUp.tsx
gurvirsasan: my bad, I meant void

client/src/components/CreatePost.tsx
gurvirsasan: Are you referring to <CssBaseline/>?

plausibly: I don't think we need <Cssbaseline>, check to make sure

PR #32, opened 2021-11-05
Update URIs of transactional emails to match routers

danielduong8:
Tested locally on Docker and also on Heroku. 

LGTM

### DIFF
--- a/server/services/emailService.ts
+++ b/server/services/emailService.ts
@@ -39,7 +39,7 @@ export default class EmailService {
     lastName: string,
     emailAddress: string
   ): Promise<boolean> {
-    const confirmURL = `${this.apiRoute}/users/confirmation/c=${confToken}`; // this will be our route
+    const confirmURL = `${this.apiRoute}/users/confirm/?c=${confToken}`; // this will be our route
     const subjectLine = "UBoard - Confirm your Email Address";
 
     const body = `Thank you for signing up to UBoard, ${firstName} ${lastName}.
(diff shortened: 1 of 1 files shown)


plausibly:
Make sure to run it and test the emails that are sent

plausibly (approved):
LGTM.

server/services/emailService.ts
```diff
@@ -39,7 +39,7 @@ export default class EmailService {
     lastName: string,
     emailAddress: string
   ): Promise<boolean> {
-    const confirmURL = `${this.apiRoute}/users/confirmation/c=${confToken}`; // this will be our route
+    const confirmURL = `${this.apiRoute}/users/confirm/?c=${confToken}`; // this will be our route
```
andy9775, line 42:
```suggestion
    const confirmURL = `${this.apiRoute}/users/confirm?c=${confToken}`; // this will be our route
```

server/services/emailService.ts
```diff
@@ -63,7 +63,7 @@ export default class EmailService {
     userName: string,
     emailAddress: string
   ): Promise<boolean> {
-    const resetURL = `${this.apiRoute}/users/password-reset/r=${confToken}`;
+    const resetURL = `${this.apiRoute}/users/reset/?r=${confToken}`;
```
andy9775, line 66:
```suggestion
    const resetURL = `${this.apiRoute}/users/password-reset?r=${confToken}`;
```

I think making it more terse here makes sense. We are resetting the password, not the user

andy9775 (approved):
Just 2 comments, otherwise LGTM!

PR #33, opened 2021-11-05
Create CRUD endpoints for posts.

andy9775:
Create the CRUD controller to:
- getPosts
- getPost (by ID)
- delete post
- update post
- upvote post
- report post

Connects up all of the controller methods to the router.

### DIFF
--- a/server/controllers/v1/post.ts
+++ b/server/controllers/v1/post.ts
@@ -1,0 +1,186 @@
+import {Post} from '../../models/post';
+
+// The maximum number of results to return.
+const MAX_RESULTS = 50;
+
+export default class PostController {
+  protected postsRepo: typeof Post;
+
+  constructor(postsRepo: typeof Post) {
+    this.postsRepo = postsRepo;
+  }
+
+  /**
+   * CRUD method to get all the posts.
+   *
+   * @param limit - Limit the posts to a number of results.
+   * @param offset - Offset the results
+   * @returns A data object containing the results.
+   */
+  async getPosts(
+    limit: number,
+    offset: number
+  ): Promise<{status: number; data: {result?: Post[]; count: number}}> {
+    const data = await this.postsRepo.findAndCountAll({
+      limit: limit > MAX_RESULTS ? MAX_RESULTS : limit,
+      // Since we are returning multiple results, we want to limit the data. This data will be shown
+      // in a list, so ignoring body is ok as it won't be displayed anyway.
+      attributes: ['id', 'title', 'feedbackScore', 'createdAt'],
+      order: [['createdAt', 'DESC']],
+      offset: offset,
+    });
+
+    return {status: data.count > 0 ? 200 : 204, data: {result: data.rows, count: data.count}};
+  }
+
+  /**
+   * Search for the post by the given ID if it exists.
+   *
+   * @param postID - The identifier used to find the specific post.
+   * @returns The details of the post
+   */
+  async getPost(
+    postID: string
+  ): Promise<{status: number; data: {result?: Post; message?: string}}> {
+    const data = await this.postsRepo.findByPk(postID);
+
+    if (!data) {
+      return {
+        status: 404,
+        data: {message: `Post ${postID} could not be found`},
+      };
+    }
+    return {
+      status: 200,
+      data: {result: data},
+    };
+  }
+
+  /**
+   * Delete the post by a given ID.
+   *
+   * @param postID - The identifier of the post to destroy.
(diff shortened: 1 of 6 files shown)


server/routes/v1/post.ts
```diff
@@ -0,0 +1,68 @@
...
+  return res.status(result.status).send(result);
+});
+
+postRouter.get('/:postid', async (req: Request, res: Response) => {
+  const result = await postController.getPost(req.params.postid);
+  res.status(result.status).send(result);
+});
+
+postRouter.delete('/:postid', async (req: Request, res: Response) => {
+  const result = await postController.deletePost(req.params.postid);
```
plausibly, line 27: Check to make sure UserID of the person attempting to delete the post matches the author

server/routes/v1/post.ts
```diff
@@ -0,0 +1,68 @@
...
+      req.body.capacity
+    );
+    res.status(result.status).send(result);
+  } catch (err) {
+    console.error(err);
+  }
+});
+
+postRouter.put('/:postid', async (req: Request, res: Response) => {
+  const result = await postController.updatePost(
```
plausibly, line 57: Similar to delete post. We should check so that other users cannot edit posts that they did not create

plausibly (changes requested):

plausibly (approved):
LGTM

server/controllers/v1/post.ts
```diff
@@ -0,0 +1,195 @@
...
+  /**
+   * Downvote a given post.
+   *
+   * @param postID - The post to downvote.
+   * @returns A status object indicating whether the post was downvoted.
+   */
+  async report(
+    postID: string
+  ): Promise<{status: number; data?: {result?: Post; message?: string}}> {
+    return this.vote(postID, -1);
```
danielduong8, line 120: Are we treating reports as simply downvotes?

If we're going to allow downvotes, I think we should either have some system in place to automatically remove a post with too many downvotes, or we should have downvoting and a separate report functionality that might be considered more seriously than a downvote. It would probably also require user authentication first. What do you guys think?

server/routes/v1/post.ts
```diff
@@ -0,0 +1,77 @@
...
+import {getAuthUser} from '../../middleware/auth';
+
+const postRouter = express.Router();
+const postController = new PostController(db.User);
+
+postRouter.get('', async (req: Request, res: Response) => {
+  const limit = req.body.limit;
+  const offset = req.body.offset;
+  if (!limit || !offset) {
+    return res.status(400).send({code: 400, message: 'Missing limit or offset'});
```
danielduong8, line 14: Sort of using you're own comment against you but, you could maybe use a ternary operator to be more specific on which is missing

server/routes/v1/post.ts
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  const result = await postController.getPosts(limit, offset);
+  return res.status(result.status).send(result);
```
danielduong8, line 18: Should we be using .json to send our result to be consistent with user routes? I believe this also sets the content-type header to json which might be better for us.

danielduong8 (commented):
Just a couple comments I could find... otherwise LGTM

server/controllers/v1/post.ts
plausibly: We can just have "like" functionality, and "report'. People can like/unlike a post and the count is visible to the public. If a post is reported, we keep track of that data in the backend (and show nothing to the frontend). A post with too many reports is deleted.

andy9775: Wouldn't report count be another field in the model?

plausibly: Yea you're right, nevermind. We can keep this, reports will downvote a post and likes will increase the score of the post. We can set a threshold where a post gets removed after being too low

danielduong8 (approved):

server/controllers/v1/post.ts
andy9775: To prevent vote spamming we also need to keep track of which users voted on what. So rather than being a score, we need another model either way. I think keeping this as-is for now makes sense, then in a future PR introducing a separate association between user and vote (userid, postid would be the two fields).

plausibly: Yes, I agree. The model can also include interactions with 'checking in' so we can prevent them from checking into an event multiple times.

andy9775: Ok. I'm merging this in so we can start connecting things up.

PR #36, opened 2021-11-09
Add author details to the Post API response.

andy9775:
Include the Post author information in the response from the get Post(s) api.

### DIFF
--- a/server/models/post.ts
+++ b/server/models/post.ts
@@ -33,6 +33,7 @@ export class Post
   UserId!: string; /* Foreign Key from UserId */
 
   static associate(model: any) {
+    Post.belongsTo(model.User);
     Post.belongsToMany(model.Tag, {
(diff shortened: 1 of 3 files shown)


danielduong8 (approved):
LGTM

server/models/post.ts
```diff
@@ -33,6 +33,7 @@ export class Post
   UserId!: string; /* Foreign Key from UserId */
 
   static associate(model: any) {
+    Post.belongsTo(model.User);
```
plausibly, line 36: User has 'hasMany' association to Post. Is this fine? I think we should check what the db looks like before merging

andy9775: This is needed.

plausibly:
LGTM

plausibly (approved):

PR #37, opened 2021-11-09
Add /me endpoint to user api.

andy9775:
/me returns the current user if authenticated and exists.

### DIFF
--- a/server/routes/v1/user.ts
+++ b/server/routes/v1/user.ts
@@ -4,6 +4,7 @@ import jwt from "jsonwebtoken";
 import EmailService from "../../services/emailService";
 import db from "../../models";
 import UserController from "../../controllers/v1/user";
+import { getAuthUser } from "../../middleware/auth";
 
 const userRouter = express.Router();
 const apiRoute = `${process.env.PAGE_URL}api/v1`;
@@ -129,11 +130,21 @@ async function resetPassHandler(req: Request, res: Response) {
   }
 }
 
+async function me(req: Request, res: Response) {
+  try {
+    const user = getAuthUser(res);
+    res.status(200).json(user);
+  } catch (err) {
+    console.error(err);
+  }
+}
+
 userRouter.post("/signin", signInHandler);
 userRouter.post("/signup", signUpHandler);
 userRouter.post("/signout", signOut);
 
 userRouter.get("/confirm", confirmEmailHandler);
 userRouter.get("/password-reset", resetPassHandler);
+userRouter.get("/me", me);
 
 export default userRouter;


server/routes/v1/user.ts
```diff
@@ -129,11 +130,21 @@ async function resetPassHandler(req: Request, res: Response) {
   }
 }
 
+async function me(req: Request, res: Response) {
+  try {
+    const user = getAuthUser(res);
+    res.status(200).json(user);
+  } catch (err) {
+    console.error(err);
```
plausibly, line 138: Respond with an error message/unauthorized?

plausibly (changes requested):

server/routes/v1/user.ts
andy9775: `getAuthUser` does.
https://github.com/GDSCUTM-CommunityProjects/UBoard/blob/main/server/middleware/auth.ts#L23

danielduong8 (approved):

server/routes/v1/user.ts
plausibly: Should we have any tests for this, since there doesnt seem to be any for getauthuser? Other than that it looks good

plausibly (approved):

server/routes/v1/user.ts
andy9775: Really this type of a test would be: 1/ set the user on locals 2/ assert that it was set (getAuthUser returns the user). So I don't really think it's worth testing a setter/getter. LMKWYT!

It is Indirectly tested in other places as well.

plausibly: Fair enough. I think it's good to merge then.

plausibly (changes requested):
I think this needs to go through the auth middleware before calling me

server/routes/v1/user.ts
```diff
@@ -129,11 +130,21 @@ async function resetPassHandler(req: Request, res: Response) {
...
+  }
+}
+
 userRouter.post("/signin", signInHandler);
 userRouter.post("/signup", signUpHandler);
 userRouter.post("/signout", signOut);
 
 userRouter.get("/confirm", confirmEmailHandler);
 userRouter.get("/password-reset", resetPassHandler);
+userRouter.get("/me", me);
```
plausibly, line 148: I think this needs to go through the auth middleware before calling me

plausibly (changes requested):

server/routes/v1/user.ts
danielduong8: ^

andy9775: #42

plausibly (approved):

PR #42, opened 2021-11-10
Enable the auth middleware.

andy9775:
Adds the ability for the auth middleware to skip certain paths and enables it application wide,
except for user authentication paths.

### DIFF
--- a/server/middleware/auth.ts
+++ b/server/middleware/auth.ts
@@ -27,8 +27,20 @@ export function getAuthUser(res: Response): User {
   return res.locals.user;
 }
 
-function auth(userRepo: typeof User) {
+/**
+ * Checks if the user is authenticated and if so, adds the user to the locals object.
+ *
+ * `skipPaths` takes a list of paths which shouldn't be checked by the auth middleware.
+ * @param userRepo Used to query for user info.
+ * @param skipPaths A list of paths which should be skipped.
+ * @returns The next handler function.
+ */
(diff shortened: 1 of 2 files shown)


plausibly (approved):
Looks good but can you test some of the routes with postman to make sure it works as intended and no minor bugs that we may have missed

andy9775:
> Looks good but can you test some of the routes with postman to make sure it works as intended and no minor bugs that we may have missed

I make sure that the skippedRoutes are actually skipped.

plausibly:
> > Looks good but can you test some of the routes with postman to make sure it works as intended and no minor bugs that we may have missed
> 
> I make sure that the skippedRoutes are actually skipped.



> > Looks good but can you test some of the routes with postman to make sure it works as intended and no minor bugs that we may have missed
> 
> I make sure that the skippedRoutes are actually skipped.

Check to make sure the regex matches valid URLs as well

andy9775:
> > > Looks good but can you test some of the routes with postman to make sure it works as intended and no minor bugs that we may have missed
> > 
> > 
> > I make sure that the skippedRoutes are actually skipped.
> 
> Check to make sure the regex properly matches valid URLs as well

I did. It'll only ignore the ones specified, otherwise everything else goes through auth.

plausibly (approved):

danielduong8 (approved):

danielduong8:
lgtm sir!

PR #43, opened 2021-11-11
Fix bugs in `router/post` and `auth`

plausibly:
Made a new PR because the previous branch was `cursed`

db.User was used in postcontroller instead of db.Post
The !offset check was preventing offset = 0
Updated auth to properly set payload of tokens (it was previously not working as intended)
Updated post controller to accept capacity of 0

### DIFF
--- a/server/routes/v1/post.ts
+++ b/server/routes/v1/post.ts
@@ -4,12 +4,12 @@ import PostController from '../../controllers/v1/post';
 import {getAuthUser} from '../../middleware/auth';
 
 const postRouter = express.Router();
-const postController = new PostController(db.User);
+const postController = new PostController(db.Post);
 
 postRouter.get('', async (req: Request, res: Response) => {
   const limit = req.body.limit;
   const offset = req.body.offset;
-  if (!limit || !offset) {
+  if (!limit || offset == undefined) {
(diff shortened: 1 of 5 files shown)


server/routes/v1/post.ts
```diff
@@ -4,12 +4,12 @@ import PostController from '../../controllers/v1/post';
...
 
 const postRouter = express.Router();
-const postController = new PostController(db.User);
+const postController = new PostController(db.Post);
 
 postRouter.get('', async (req: Request, res: Response) => {
   const limit = req.body.limit;
   const offset = req.body.offset;
-  if (!limit || !offset) {
+  if (!limit || offset == undefined) {
```
andy9775, line 12: Limit can also technically be 0 which would (correctly) return an empty array. 

I'll leave it up to you if you want to display an error for a 0 limit, or just allow it to return an empty array.

andy9775 (approved):
Just one comment, otherwise LGTM!

server/routes/v1/post.ts
plausibly: Limit of 0 doesn't really make sense... We can handle the errors to return empty on the frontend

PR #46, opened 2021-11-11
Connect dashboard to API

plausibly:
- Periodically fetch post data from the API and load it onto dashboard
- "Read more" does another API request which fetches the rest of the post data
- Delete button is functional
- Added UserContext which contains the current logged in user
- Dashboard forces you to login if you are not properly authorized
- Sign Out button integrated with API
- Delete/Edit only show up if the user is authorized
- Style / Bug fixes

Notes: Need to update POST type after Daniel's PR is merged.


andy9775:
I'm confused as to why this was reopened.

plausibly:
> I'm confused as to why this was reopened.

I was updating some things. We did not merge it to main branch

client/src/components/CreatePost.tsx
```diff
@@ -37,14 +53,31 @@ function CreatePost() {
...
   return (
+    <>
+      <Button
+        variant="outlined"
+        onClick={() => {
+          toggleDialog(true);
+        }}
+        data-testid="btn-newpost"
+      >
+        New Post
```
andy9775, line 71: `Create a Post`?

client/src/containers/PostDashboard.tsx
```diff
@@ -1,13 +1,49 @@
...
+      })
+      .then((res: any) => {
+        if (res.data.data.result) {
+          updateRecent(res.data.data.result);
+        }
+      })
+      .catch((err) => console.log(err));
+  };
+
+  React.useEffect(() => {
```
andy9775, line 31: We could store the posts in app wide state, set the polling logic in app.tsx, then this way we will always be polling not just when the user is on the posts dashboard.

andy9775 (changes requested):
I noticed some inconsistencies with double/single quotes. Let's make sure we stick to one. A note on deleting posts (and things in general limited to a specific user). Hit the /me endpoint to get user info, then based on the uid of the post and the uid of the current user, we can enable/disable certain things. We should use something like redux or context to store the state of the current user so we only have to fetch it once. 

Otherwise, request a review once you update the client api stuff.

client/src/containers/PostDashboard.tsx
plausibly: The site does always poll. Since "View post" and "New post" are in dialogs, post dashboard is technically open in the background.

client/src/components/CreatePost.tsx
plausibly: I'll actually revert this file to the state of GDSC main since Gurvir has this in his PR

danielduong8:
comment on my PR so that i ermember to integrate sign out

plausibly:
I updated the PR message at the top to reflect my changes

client/src/containers/PostDashboard.tsx
```diff
@@ -46,14 +57,46 @@ function RecentPosts() {
...
 export default function PostDashboard() {
+  const [currentUser, setUser] = React.useState(null);
+  const [isLoading, setLoading] = React.useState(true);
+  const navigate = useNavigate();
+
+  const authBarrier = () => {
+    // force a login
+    navigate('/');
+    setUser(null);
+  };
```
andy9775, line 68: We could to something like this using react router, this way it automatically does checks on all pages.

client/src/components/ViewPostDialog.tsx
```diff
@@ -183,26 +201,39 @@ export default function ViewPostDialog(props: {
...
-        if (res.data) {
-          setData(res.data.data.result);
+    api
+      .fetchPost(props.postUser.id)
+      .then(
+        (
+          res: APIResponse<{ data: { result?: PostUser }; message?: string }>
+        ) => {
+          if (res.data && res.data.data && res.data.data.result) {
+            // appeasing TS
```
andy9775, line 216:
```suggestion

```
Doesn't really add value

andy9775 (changes requested):
Some first pass comments.

Can you undo the formatting changes for now? And only have them in the formatting PR? You can do this by having save on format, format only the changed lines and not the whole file.

I would also add a test which checks for the post being created. Mock the `api` class and assert that posts are created when the data is valid. Then, mock the return and make sure it's being displayed correctly. Same for pagination, mock a list of fake data being returned and make sure that the first and last elements for  example are being updated.

client/src/containers/PostDashboard.tsx
```diff
@@ -5,26 +5,37 @@ import Container from '@mui/material/Container';
...
-
-import { api_v1 } from '../api/v1';
 import CreatePost from '../components/CreatePost';
 
+import ServerApi, { APIResponse, PostUser } from '../api/v1';
+import { useNavigate } from 'react-router-dom';
+
+const api = new ServerApi();
+
+export const UserContext: React.Context<{
```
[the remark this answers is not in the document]
plausibly: But the user will not be authenticated on App.tsx, we only want to get the user after authentication (which will route them to the dashboard). The rest of our site is contained on the dashboard

client/src/api/v1/index.ts
```diff
@@ -1,4 +1,101 @@
-import axios from 'axios';
+import axios, { AxiosInstance } from 'axios';
+import { Post } from 'models/post';
+import { User } from 'models/user';
 
-export const api_v1 = axios.create({ baseURL: `api/v1` });
-// TODO: This will be changed to Daniel's API class
+export type PostUser = Post & {
```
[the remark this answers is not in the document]
plausibly: I think this is fine for now, we dont want to make the controller class publicly available.

plausibly:
I wasn't able to fix the bug where the currently viewed post would change if a new post was made in the background. Instead, when a user is viewing a post the dashboard will not poll data. I also updated useEffect to fetch instantly once they close out of the post, and then resume polling.

client/src/containers/PostDashboard.tsx
plausibly: How do you mean? Like disabling the "/dashboard" route on App.tsx until they are authed?

client/src/components/PostPreview.tsx
```diff
@@ -10,13 +10,45 @@ import Typography from "@mui/material/Typography";
...
 
+  /**
+   * Return a string of the date relative to now in minutes, hours or days. If the date
+   * is over a month, return the date stringified. 
+   * 
+   * @param base - The base date being compared with
+   * @param date - The date converted into relative time
+   * @returns String containing the date relative to now
+   */
+function relativeTime(date: Date) {
```
andy9775, line 23: There should be libraries which do this for us. Benefit would be that all cases are handled and they are tested.

client/src/components/ViewPostDialog.tsx
```diff
@@ -70,11 +69,23 @@ function MoreOptions(props: {
...
+          setMsg('Post has been reported.');
+        } else {
+          setMsg('Failed to report post. ');
+        }
+        showAlert(true);
+      })
+      .catch(() => {
+        setMsg('Failed to report post.');
+        showAlert(true);
+      });
```
andy9775, line 87:
```suggestion
    api
      .reportPost(props.postID)
      .then((res) => {
        if (res.status === 204) {
          setMsg('Post has been reported.');
        } else {
          setMsg('Failed to report post. ');
        }
      })
      .catch(() => {
        setMsg('Failed to report post.');
      }).finally(() => showAlert(true));
```

client/src/containers/PostDashboard.tsx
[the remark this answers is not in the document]
andy9775: It could be an empty object. The typical pattern is to have user data in something like redux which is app wide state.

I feel like there would be a need to make a change in the future, but I'll leave it up to you.

client/src/containers/PostDashboard.tsx
```diff
@@ -11,46 +11,68 @@ import Header from '../components/Header';
 import PostPreview from '../components/PostPreview';
 import CreatePost from '../components/CreatePost';
 
-import ServerApi, { APIResponse, PostUser } from '../api/v1';
+import ServerApi, { PostUserPreview } from '../api/v1';
 import { useNavigate } from 'react-router-dom';
```
andy9775:
```suggestion
import { useNavigate } from 'react-router-dom';

import ServerApi, { PostUserPreview } from '../api/v1';
```

client/src/containers/PostDashboard.tsx
```diff
@@ -11,46 +11,68 @@ import Header from '../components/Header';
...
 
+  /* Fetch posts triggered by page-change or post dialog close */
+  React.useEffect(() => {
+    if (!openedPost) {
+      api
+        .fetchRecentPosts(POSTS_PER_PAGE, POSTS_PER_PAGE * (props.pageNum - 1))
+        .then((res) => {
+          if (res.data && res.data.data.result) {
+            updateRecent(res.data.data.result);
+            props.setPageCount(Math.ceil(res.data.data.count / POSTS_PER_PAGE));
```
andy9775, line 65: I would write some tests for the pagination logic to make sure the correct page is being displayed and the correct data is being displayed. You can just create some mock data to be returned from the mock `fetchRecentPosts` method and just assert that the UI displays correctly.

client/src/containers/PostDashboard.tsx
```diff
@@ -141,7 +174,7 @@ export default function PostDashboard() {
             color='text.secondary'
             component='p'
           >
-            You're all up to date!
+            You"re all up to date!
```
andy9775, line 177:
```suggestion
            You're all up to date!
```

client/src/containers/PostDashboard.tsx
```diff
@@ -125,7 +150,15 @@ export default function PostDashboard() {
             alignItems: 'center',
           }}
         >
-          <Pagination count={1} color='primary' variant='outlined' />
+          <Pagination
+            count={pageCount}
+            page={page}
+            onChange={(event: React.ChangeEvent<unknown>, pg: number) => {
```
andy9775, line 156:
```suggestion
            onChange={(event: React.ChangeEvent<{}>, pg: number) => {
```

server/controllers/v1/post.ts
```diff
@@ -26,25 +36,34 @@ export default class PostController {
...
 
     return {
-      status: data.count > 0 ? 200 : 204,
-      data: { result: data.rows as PostUser[], count: data.count },
+      status: data[0].count > 0 ? 200 : 204,
+      data: {
+        result: data[0].rows as any as PostUserPreview[],
+        count: data[0].count,
+        total: data[1],
+      },
```
andy9775, line 66: Can you write a test for this to ensure that the data is correct?

andy9775 (changes requested):
If deciding to keep the `relativeTime` function, make sure you have proper tests, it seems edge case-y.

I would also add some test for the pagination stuff. Next review should primarily be that.

client/src/containers/PostDashboard.tsx
plausibly: I have to explicitly put unknown or it causes an error, I looked into this on the MUI docs and they also put unknown for this.

``Type 'ChangeEvent<unknown>' is not assignable to type 'ChangeEvent<{}>'.``

server/controllers/v1/post.ts
plausibly: sure

client/src/components/PostPreview.tsx
plausibly: I added some tests for this

client/src/api/v1/index.ts
```diff
@@ -13,10 +13,10 @@ export type PostUserPreview = {
   title: string;
   createdAt: string;
 } & {
-  User: { id: string; firstName: string; lastName: string; userName: string };
+  User: { id: string; firstName: string; lastName: string };
 };
 
-export class ApiError extends Error {}
+class ApiError extends Error {}
```
andy9775, line 19: This error should be caught and handled.

client/src/containers/PostDashboard.test.tsx
```diff
@@ -54,6 +55,196 @@ describe("Test interaction with dashboard", () => {
...
+      .mockImplementation((limit, offset) => {
+        const results = posts.slice(offset, limit + offset);
+        return Promise.resolve({
+          status: 200,
+          data: {
+            data: {
+              result: results,
+              count: results.length,
+              total: posts.length,
+              message: 'asd',
```
andy9775, line 215:
```suggestion
```
Message is optional, and is only returned if an error occurs.

andy9775 (approved):
LGTM!

client/src/containers/PostDashboard.tsx
andy9775: https://ui.dev/react-router-v5-protected-routes-authentication/

On loading a route you can check if the user object exists in app state. If not, redirect out to the login page, otherwise display the component for that page. 

```typescript
export function ProtectedRoute(props: {component: React.Component}) {
    const [user, ...] = getUserFromContext(); // some logic here
   if (user & user.isAuthenticated) {
     return <component/>
  }
  redirectToLoginPage(); // you get it
}
```

It's much safer to have it centralized in one location

client/src/App.tsx
```diff
@@ -1,33 +1,83 @@
...
-          {/* TODO: Ensure authenticated then redirect to dashboard */}
-        </Routes>
-      </BrowserRouter>
-    </ThemeProvider>
+      <ThemeProvider theme={theme}>
+        <CssBaseline />
+        <BrowserRouter>
+          <Routes>
+            <Route path='/' element={<AuthContainer setAuthed={setAuthed}/>} />
+            <Route path='/dashboard' element={<ProtectedRoute isAuth={authed}/>} />
```
danielduong8, line 75: I think in cases like these it's probably a good idea to use contexts rather than passing down a state hook.

danielduong8 (approved):

PR #47, opened 2021-11-11
Connect CreatePost to API

gurvirsasan:
- Implemented a dialog 
- back button closes the dialog
- Create button is connected to the api
- added some alerts for catch and api issues
- added tests for the back button functionality

### DIFF
--- a/client/src/api/v1/index.ts
+++ b/client/src/api/v1/index.ts
@@ -126,6 +126,20 @@ export default class ServerApi {
       {},
       { data: { result?: PostUser }; message?: string }
     >(`/posts/${postID}`, {});
+    }
+
+  async createPost(form: {
+    title: string;
+    body: string;
+    file: string;
+    tags: string;
+    capacity: string;
+    location: string;
+  }) {
+    return await this.post<typeof form, { result?: Post; message?: string }>(
+      '/posts/',
+      form
+    );
   }
 
   async signUp(form: {
--- a/client/src/components/CreatePost.test.tsx
+++ b/client/src/components/CreatePost.test.tsx
@@ -1,18 +1,18 @@
-import React from "react";
-import CreatePost from "./CreatePost";
-import { unmountComponentAtNode } from "react-dom";
-import { act } from "react-dom/test-utils";
-import { render, screen, fireEvent, cleanup } from "@testing-library/react";
-import newImage from "../assets/background.jpg";
+import React from 'react';
+import CreatePost from './CreatePost';
+import { unmountComponentAtNode } from 'react-dom';
+import { act } from 'react-dom/test-utils';
+import { render, screen, fireEvent, cleanup } from '@testing-library/react';
 
 let container: HTMLElement | null = null;
 beforeEach(() => {
   // setup a DOM element as a render target
-  container = document.createElement("div");
+  container = document.createElement('div');
   document.body.appendChild(container);
   act(() => {
     render(<CreatePost />);
   });
+  screen.getByTestId('newPostButton').click();
 });
 
 afterEach(() => {
@@ -25,34 +25,38 @@ afterEach(() => {
   }
 });
 
-describe("verifying launch of create post component", () => {
-  it("shows the preview button as disabled", () => {
+describe('verifying launch of create post component', () => {
+  it('shows the preview button as disabled', () => {
     // get the preview button
-    expect(screen.getByTestId("previewButton")).toBeDisabled();
+    expect(screen.getByTestId('previewButton')).toBeDisabled();
   });
 
-  it("should enable the preview button after providing required fields", () => {
+  it('should enable the preview button after providing required fields', () => {
     // input title
-    const titleTextField = screen.getByPlaceholderText("title");
-    fireEvent.change(titleTextField, { target: { value: "Test Club" } });
+    const titleTextField = screen.getByPlaceholderText('title');
+    fireEvent.change(titleTextField, { target: { value: 'Test Club' } });
     // input body
-    const bodyTextField = screen.getByPlaceholderText("description");
-    fireEvent.change(bodyTextField, { target: { value: "body" } });
+    const bodyTextField = screen.getByPlaceholderText('description');
+    fireEvent.change(bodyTextField, { target: { value: 'body' } });
     fireEvent.blur(bodyTextField);
-    expect(screen.getByTestId("previewButton")).not.toBeDisabled();
+    expect(screen.getByTestId('previewButton')).not.toBeDisabled();
   });
 
-  it("renders `previewPopUp` component upon clicking Preview", () => {
+  it('renders `previewPopUp` component upon clicking Preview', () => {
     // input title
-    fireEvent.change(screen.getByPlaceholderText("title"), {
-      target: { value: "Test Club" },
+    fireEvent.change(screen.getByPlaceholderText('title'), {
+      target: { value: 'Test Club' },
     });
     // input body
-    const bodyTextField = screen.getByPlaceholderText("description");
-    fireEvent.change(bodyTextField, { target: { value: "body" } });
+    const bodyTextField = screen.getByPlaceholderText('description');
+    fireEvent.change(bodyTextField, { target: { value: 'body' } });
     fireEvent.blur(bodyTextField);
     // pop up should render
-    screen.getByTestId("previewButton").click();
-    expect(screen.getByTestId("PreviewPopUpComponent")).toBeInTheDocument();
+    screen.getByTestId('previewButton').click();
+    expect(screen.getByTestId('PreviewPopUpComponent')).toBeInTheDocument();
+  });
+
+  it('closes the dialog when clicked on `Back` button', () => {
+    expect(screen.getByTestId('newPostButton')).toBeInTheDocument();
   });
 });
--- a/client/src/components/CreatePost.tsx
+++ b/client/src/components/CreatePost.tsx
@@ -1,180 +1,260 @@
-import React, { useState } from "react";
+import React, { useState } from 'react';
 
-import Grid from "@mui/material/Grid";
-import Box from "@mui/material/Box";
-import Container from "@mui/material/Container";
-import Button from "@mui/material/Button";
-import TextField from "@mui/material/TextField";
-import Divider from "@mui/material/Divider";
-import Typography from "@mui/material/Typography";
-import Tooltip from "@mui/material/Tooltip";
+import Grid from '@mui/material/Grid';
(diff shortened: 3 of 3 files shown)


client/src/components/CreatePost.test.tsx
```diff
@@ -13,6 +12,9 @@ beforeEach(() => {
   act(() => {
     render(<CreatePost />);
   });
+  // const newPostButton = screen.findByTestId("newPostButton");
```
plausibly, line 15: Remove commented code

client/src/api/v1/index.ts
```diff
@@ -0,0 +1,3 @@
+import axios from "axios";
```
plausibly, line 1: Replace with class (check daniels PR)

plausibly (changes requested):

client/src/components/CreatePost.test.tsx
```diff
@@ -3,7 +3,6 @@ import CreatePost from "./CreatePost";
 import { unmountComponentAtNode } from "react-dom";
 import { act } from "react-dom/test-utils";
 import { render, screen, fireEvent, cleanup } from "@testing-library/react";
-import newImage from "../assets/background.jpg";
 
 let container: HTMLElement | null = null;
```
danielduong8, line 7: I don't believe this as well as the afterEach are necessary for the tests to work, you can remove them.

client/src/components/CreatePost.tsx
```diff
@@ -8,22 +8,39 @@ import TextField from "@mui/material/TextField";
...
+    location: "",
   });
 
   const [openPopup, setOpenPopup] = useState(false);
 
+  const handleSubmit = () => {
+    api_v1
+      .post("/posts/", form)
+      .then((res) => {
+        console.dir(res);
```
danielduong8, line 35: I don't think it's a good idea to log an entire res object to the console where its plainly visible to anyone who opens a dev console. I think you just need to handle the data in res, you don't necessarily need to log it to the console.

danielduong8 (changes requested):

client/src/components/CreatePost.tsx
andy9775: Agreed. But this should change with the clientApi class either way.

client/src/components/CreatePost.tsx
```diff
@@ -37,144 +54,185 @@ function CreatePost() {
...
-              <Tooltip title="Enter all required fields!">
-                <Box>
+            <Typography
+              component="h3"
+              variant="h3"
+              fontWeight="bold"
+              paddingTop="10px"
+            >
+              Create Post
+            </Typography>
```
andy9775, line 104: Can this not just be an `<h3>...</h3>`?

client/src/components/CreatePost.test.tsx
```diff
@@ -55,4 +57,9 @@ describe("verifying launch of create post component", () => {
     screen.getByTestId("previewButton").click();
     expect(screen.getByTestId("PreviewPopUpComponent")).toBeInTheDocument();
   });
+
+  it("closes the dialog when clicked on `Back` button", () => {
+    screen.getByTestId("backButton").click();
+    expect(screen.getByTestId("newPostButton")).toBeInTheDocument();
```
andy9775, line 63: I think with a dialog, since it sits on top of something, the elements in the background are still in the document. Verity this by commenting out line 62 and verifying that the test fails (you don't need a specific test for this, just check whether my assumption is correct).

andy9775 (changes requested):
Wait for @danielduong8 to merge his stuff in and follow up with the api client class. Tests should assert that the `createPost` function was called with the correct params and such. Follow his tests, and logic.

client/src/components/CreatePost.tsx
danielduong8: I think its fine right? Like in the case we might want to apply a theme to this? Unless theming still works with plain html tags, I'm not too sure.

andy9775: It does. Two options 1/ useTheme and apply via the style attribute 2/ use styled-components. It's an extra dependency and import that isn't *really* necessary. 

It's OK to leave it in, I'm not going to be picky about it, but I would just use a regular HTML tag.

client/src/components/CreatePost.tsx
```diff
@@ -37,144 +54,185 @@ function CreatePost() {
...
+                </Grid>
+
+                <Grid item xs={12} md={3}>
+                  <TextField
+                    label="Event Capacity"
+                    placeholder="40"
+                    fullWidth
+                    size="small"
+                    onChange={(e) =>
+                      setForm({ ...form, capacity: e.target.value })
```
plausibly, line 155: Can we add a check to make sure capacity is numeric?

plausibly (changes requested):

client/src/components/CreatePost.test.tsx
gurvirsasan: Your assumption is correct! Tried it and it still worked.

client/src/api/v1/index.ts
```diff
@@ -18,6 +18,10 @@ export default class ServerApi {
     }
   }
 
+  async createPost(path: string, form: {}) {
+    return await this.post(path, form);
```
plausibly, line 22: Don't need `path` here, just put `/posts/`

client/src/components/CreatePost.tsx
```diff
@@ -8,173 +8,251 @@ import TextField from "@mui/material/TextField";
...
+    toggleDialog(false);
+  };
   const handleClickOpen = () => {
     setOpenPopup(true);
   };
 
-  // handle functions
+  // handling
+  const handleSubmit = () => {
+    API.createPost(`/posts/`, form)
```
plausibly, line 45: After you change the function, just call it with API.createPost(form)

client/src/components/CreatePost.tsx
```diff
@@ -8,173 +8,251 @@ import TextField from "@mui/material/TextField";
...
 import Tooltip from "@mui/material/Tooltip";
-
+import Dialog from "@mui/material/Dialog";
+import { ArrowBack } from "@mui/icons-material";
 import PreviewPopUp from "./PreviewPopUp";
+import Snackbar from "@mui/material/Snackbar";
+
+import ServerApi from "../api/v1/index";
+
+const API = new ServerApi();
```
andy9775, line 18: I think making this lower case would be better

plausibly (changes requested):

client/src/components/CreatePost.tsx
```diff
@@ -8,173 +8,251 @@ import TextField from "@mui/material/TextField";
...
     tags: "",
-    eventCapacity: "",
-    eventLocation: "",
+    capacity: "",
+    location: "",
   });
+  const [openPopup, setOpenPopup] = useState(false); // for preview popup
+  const [isAlertOpen, showAlert] = useState(false); // for snackbar
+  const [alertMsg, setMsg] = useState("An error has occurred"); // for snackbar message
+  const [capacityError, setCapacityError] = React.useState(""); // for capacity input validation
```
andy9775, line 33:
```suggestion
  const [capacityError, setCapacityError] = useState(""); // for capacity input validation
```

client/src/components/CreatePost.tsx
```diff
@@ -8,173 +8,251 @@ import TextField from "@mui/material/TextField";
...
+          setMsg("Post has been succesfully created.");
+        } else {
+          setMsg("Failed to create post");
+        }
+        showAlert(true);
+      })
+      .catch((err) => {
+        console.error(err);
+        setMsg("Failed to create post. Ensure all the fields are correct");
+        showAlert(true);
```
andy9775, line 57: since show alert is executed in both cases, you can put it into a `finally`
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/finally

client/src/api/v1/index.ts
```diff
@@ -7,28 +11,55 @@ export default class ServerApi {
...
 
+  async createPost(form: {
+    title: string;
+    body: string;
+    file: string;
+    tags: string;
+    capacity: string;
+    location: string;
+  }) {
+    return await this.post<typeof form, { result?: Post; message?: string }>(
```
plausibly, line 40: I think the response type for this is: ``data: { result?: Post; message?: string }`` and not ``{ result?: Post; message?: string }``... just check to make sure.

When I was working on my API calls it was the former. In that case, you'll need to update any API calls that use the response data to be res.data.data.result instead of res.data.result

client/src/components/CreatePost.tsx
```diff
@@ -1,180 +1,260 @@
...
+              body={form.body}
+              img={form.file}
+              tags={form.tags}
+              eventCapacity={form.capacity}
+              location={form.location}
+              openPopup={openPopup}
+              handleClose={() => setOpenPopup(false)}
+            ></PreviewPopUp>
+          </Box>
+        </Container>{' '}
```
plausibly, line 255: ??? ``{' '}``

plausibly (changes requested):

client/src/components/CreatePost.tsx
```diff
@@ -1,180 +1,260 @@
...
     setOpenPopup(true);
   };
 
-  // handle functions
+  // handling
+  const handleSubmit = () => {
+    api
+      .createPost(form)
+      .then((res) => {
+        if (res.status === 204) {
```
plausibly, line 48: 200

client/src/components/CreatePost.tsx
```diff
@@ -1,180 +1,260 @@
...
+  const [openPopup, setOpenPopup] = useState(false); // for preview popup
+  const [isAlertOpen, showAlert] = useState(false); // for snackbar
+  const [alertMsg, setMsg] = useState('An error has occurred'); // for snackbar message
+  const [capacityError, setCapacityError] = useState(''); // for capacity input validation
+  const [isOpen, toggleDialog] = useState(false); // for create post dialog toggle
 
-  const [openPopup, setOpenPopup] = useState(false);
-
-  // handle functions
+  const closeDialog = () => {
```
plausibly, line 36: In closeDialog (before toggleDialog) empty the form (i.e set title, body, ..., to empty string)

plausibly (changes requested):
See above comments.

Additionally, in PreviewPopup remove this typography (line 55-57)
```
            <Typography sx={{ fontStyle: "italic" }} display="inline">
              x mins ago by UserName
            </Typography>
```

PR #48, opened 2021-11-12
Integrate api with login page

danielduong8:
Refactor state hooks and integrate api 

- Refactored state hooks for cleaner code

- Add form validation before calling api

- Install axios

Setup navigation on successful login

Add api class

Add MaterialUI Alert components


client/src/api/v1/index.tsx
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  async signIn(
+    form: { userName: string; password: string },
+    navigate: NavigateFunction,
+    from: string
+  ) {
+    try {
+      await this.api.post("/users/signin", form);
+      navigate(from, { replace: true });
```
andy9775, line 37: You don't want to navigate here. This should only be network related stuff.

client/src/containers/AuthContainer.test.tsx
```diff
@@ -98,3 +88,122 @@ describe("verifying valid input for signup page", () => {
     });
   });
 });
+
+describe("signUp component", () => {
+  describe("handleSubmit function", () => {
```
andy9775, line 93: Better test name

client/src/containers/AuthContainer.test.tsx
```diff
@@ -98,3 +88,122 @@ describe("verifying valid input for signup page", () => {
...
+
+    it("should show email confirmation alert on success", async () => {
+      screen.getByTestId("SignUpTabButton").click();
+
+      const form = screen.getByTestId("form");
+      const mockSignUp = jest
+        .spyOn(ServerApi.prototype, "signUp")
+        .mockImplementation(() =>
+          Promise.resolve({ response: {} as AxiosResponse })
+        );
```
andy9775, line 156: This is perfect. What you could also do, 1/ is assert that it was called with the expected parameters. 2/ return some mock data then assert that is is displayed in the UI as expected (this is more relevant for the `posts` logic, but I'm leaving it here for reference).

client/src/containers/AuthContainer.test.tsx
```diff
@@ -98,3 +88,122 @@ describe("verifying valid input for signup page", () => {
...
+      const mockAlert = jest
+        .spyOn(window, "alert")
+        .mockImplementation(() => {});
+
+      fireEvent.submit(form);
+      await new Promise((r) => setTimeout(r, 1));
+      expect(mockSignUp).toBeCalled();
+      expect(mockAlert).toBeCalledWith("Email confirmation sent!");
```
andy9775, line 164: If you assert that mockSignup was called with certain params (the expected ones) you can avoid mockAlert and asserting that it was triggered. Testing the call params is more robust.

client/src/containers/AuthContainer.test.tsx
```diff
@@ -98,3 +88,122 @@ describe("verifying valid input for signup page", () => {
...
+    });
+  });
+});
+
+describe("Login component", () => {
```
andy9775, line 169: Successful login test?

client/src/containers/AuthContainer.test.tsx
```diff
@@ -98,3 +88,122 @@ describe("verifying valid input for signup page", () => {
...
+        );
+      });
+
+      const form = screen.getByTestId("form");
+      const mockAlert = jest
+        .spyOn(window, "alert")
+        .mockImplementation(() => {});
+
+      fireEvent.submit(form);
+      expect(mockAlert).toBeCalledWith("Passwords do not match.");
```
andy9775, line 116: You can assert that the mockApi function was not called rather than asserting that the alert was called with some error. If the error message changes for example, that would cause the test to fail when it really shouldn't (since there error message is just a visual thing, whereas this is a logic test).

andy9775 (changes requested):
I don't think the api client tests are really necessary. I would also avoid `window.error`. 
Otherwise good job, some some minor fixes/comments.

client/src/containers/AuthContainer.test.tsx
danielduong8: How would we test the different params since we can't actually test what happens to mockSignUp based on these different inputs?

Also, I'm not sure how to differentiate a successful sign up from a failed signup without checking the alert messages logged.

andy9775: `expect(mockSignUp).toBeCalledWith(expectedEmail, expectedUserName, ...)` where expectedEmail is the email that the form input was set to. So if in your test, you set the users email in the form to be 'foo@mail.utoronto.ca`, the mock signup function should be called with  'foo@mail.utoronto.ca` as the first param (similar logic extends to username, password, passwordConfirmation, first and last name). If the mock signup function has not been called, that means that some component level validation prevented that from happening (e.g. the email format was bad). 

So if the mock signup function was called, that means that it was a successful signup. If there was not successful signup, e.g. email format was invalid, or the password was too short, then the mockSignup function should not be called.

danielduong8:
The last commit is just renaming a bunch of "login" to "signin"; you can just skip that commit and look at the previous one to see the actual changes.

I'm sorry I just had to lol, it was bothering me greatly.

client/src/components/SignIn.tsx
```diff
@@ -8,32 +8,52 @@ import Box from "@mui/material/Box";
...
     password: "",
   });
 
-  const [errors, setErrors] = useState({
+  const [signinFormErrors, setSignInFormErrors] = useState({
     userName: "",
     password: "",
   });
 
+  const formIsMissingFields = () => {
```
andy9775, line 38: isFormMissingFields

client/src/components/SignUp.tsx
```diff
@@ -10,76 +10,106 @@ import Typography from "@mui/material/Typography";
...
+    firstName: "",
+    lastName: "",
     email: "",
     userName: "",
     password: "",
-    firstName: "",
-    lastName: "",
   });
 
+  const formIsMissingFields = () => {
```
andy9775, line 33: isFormMissingFields

client/src/containers/AuthContainer.test.tsx
```diff
@@ -91,119 +89,155 @@ describe("verifying valid input for signup page", () => {
...
+          password: "password",
+          firstName: "Daniel",
+          lastName: "Zingaro",
+        };
+
+        await act(async () => {
+          fireEvent.submit(form);
+          await new Promise((r) => setTimeout(r, 1));
+        });
+        expect(mockSignUp).toBeCalledWith(formData);
```
andy9775, line 165: Perfect

client/src/containers/AuthContainer.test.tsx
```diff
@@ -91,119 +89,155 @@ describe("verifying valid input for signup page", () => {
...
+
+      });
     });
   });
 });
 
-describe("Login component", () => {
+describe("SignIn component", () => {
   describe("handleSubmit function", () => {
```
andy9775, line 173: I would rename this to be something more descriptive of the process being tested, not some inner function name while we're here.

client/src/containers/AuthContainer.test.tsx
```diff
@@ -91,119 +89,155 @@ describe("verifying valid input for signup page", () => {
...
+        expect(mockSignIn).not.toBeCalled();
+      });
 
-      const form = screen.getByTestId("form");
-      const mockSignIn = jest
-        .spyOn(ServerApi.prototype, "signIn")
-        .mockImplementation(() =>
-          Promise.resolve({ response: {} as AxiosResponse })
-        );
+      it("should not call SignIn if errors are present", async () => {
```
andy9775, line 187: I would add a test for a backend response indicating a bad password or something

client/src/api/v1/index.ts
```diff
@@ -7,7 +7,7 @@ export default class ServerApi {
     this.api = axios.create({ baseURL: `api/v1` });
   }
 
-  async post(path: string, form: any) {
+  protected async post<Type>(path: string, form: Type) {
```
andy9775, line 10: I would make the return type a generic as well. E.g. sign-in IIRC returns some user info.

https://www.oreilly.com/library/view/learning-typescript-2x/9781788391474/b170884f-368a-417e-9ba4-b5e86207b8dd.xhtml

andy9775 (approved):
I noticed that your taking an alert function as a prop, you coud combine it with https://reactjs.org/docs/portals.html you could just have a set of functions you import: `alertSuccess(msg)`, `alertError(msg)` and so on.  Alternatively, you could use https://reactjs.org/docs/context.html, have the context provider and alert component at the root (App.tsx) and then each  sub-component can just useContext (https://reactjs.org/docs/context.html) and you could have a bunch of alert functions passed in that way. This would be much cleaner and more reusable across different components, but I'll leave it up to you if you want to go that route. 

Just a few comments, otherwise LGTM! Feel free to submit once you address them

plausibly (approved):

plausibly (changes requested):
Could you setup the "Logout" button on the post dashboard?

Daniel: Yes but probably in a separate PR.

PR #62, opened 2021-11-25
Create post api

gurvirsasan:
Accidentally closed #47 

- Implemented a dialog 
- back button closes the dialog
- Create button is connected to the api
- added some alerts for catch and api issues
- added tests for the back button functionality

### DIFF
--- a/client/src/components/CreatePost.test.tsx
+++ b/client/src/components/CreatePost.test.tsx
@@ -36,9 +36,7 @@ describe('verifying launch of create post component', () => {
     const titleTextField = screen.getByPlaceholderText('title');
     fireEvent.change(titleTextField, { target: { value: 'Test Club' } });
     // input body
-    const bodyTextField = screen.getByPlaceholderText(
-      'Description (minimum 25 characters)'
-    );
+    const bodyTextField = screen.getByTestId('bodyTextField');
     fireEvent.change(bodyTextField, {
       target: { value: 'This should be at least 25 characters long' },
     });
@@ -52,9 +50,7 @@ describe('verifying launch of create post component', () => {
       target: { value: 'Test Club' },
     });
     // input body
-    const bodyTextField = screen.getByPlaceholderText(
-      'Description (minimum 25 characters)'
-    );
+    const bodyTextField = screen.getByTestId('bodyTextField');
     fireEvent.change(bodyTextField, {
       target: { value: 'This should be at least 25 characters long' },
     });
--- a/client/src/components/CreatePost.test.tsx
+++ b/client/src/components/CreatePost.test.tsx
@@ -36,8 +36,12 @@ describe('verifying launch of create post component', () => {
     const titleTextField = screen.getByPlaceholderText('title');
     fireEvent.change(titleTextField, { target: { value: 'Test Club' } });
     // input body
-    const bodyTextField = screen.getByPlaceholderText('description');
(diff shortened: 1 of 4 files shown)


plausibly (approved):

client/src/components/CreatePost.test.tsx
```diff
@@ -25,34 +25,46 @@ afterEach(() => {
...
     // input title
-    const titleTextField = screen.getByPlaceholderText("title");
-    fireEvent.change(titleTextField, { target: { value: "Test Club" } });
+    const titleTextField = screen.getByPlaceholderText('title');
+    fireEvent.change(titleTextField, { target: { value: 'Test Club' } });
     // input body
-    const bodyTextField = screen.getByPlaceholderText("description");
-    fireEvent.change(bodyTextField, { target: { value: "body" } });
+    const bodyTextField = screen.getByPlaceholderText(
+      'Description (minimum 25 characters)'
```
andy9775, line 40: The placeholder text tends to change often. I would use something like test id, id or some other attribute here.

goes for other uses of `getByPlaceholderText` as well

client/src/components/CreatePost.tsx
```diff
@@ -1,181 +1,281 @@
...
-    body: "",
-    file: "",
-    tags: "",
-    eventCapacity: "",
-    eventLocation: "",
+    title: '',
+    body: '',
+    file: '',
+    tags: '',
+    capacity: '',
```
andy9775, line 27: Shouldn't capacity be an integer? Default can be -1 or something for an unset capacity.

client/src/components/CreatePost.tsx
```diff
@@ -1,181 +1,281 @@
...
+      location: '',
+    });
+    toggleDialog(false);
+  };
   const handleClickOpen = () => {
     setOpenPopup(true);
   };
 
-  // handle functions
+  // handling
```
andy9775, line 51:
```suggestion
```

Comments should generally explain why you're doing something a certain way if it's not clear, not what it does (unless it's super unreadable).

andy9775 (approved):
WRT to the tests, I would add one where you mock the createPost method of the api class and then assert that it is called when it needs to be called, and not called when it shouldn't be.

Otherwise LGTM!

client/src/components/CreatePost.test.tsx
gurvirsasan: I tried doing it with test id however it kept failing.
Turns out its because `TextField` is an abstraction by Material ui which was causing the issue.

client/src/components/CreatePost.tsx
```diff
@@ -1,181 +1,279 @@
...
-  const [openPopup, setOpenPopup] = useState(false);
-
-  // handle functions
+  const closeDialog = () => {
+    setForm({
+      title: '',
+      body: '',
+      file: '',
+      tags: '',
+      capacity: -1,
```
plausibly, line 41: set default capacity to 0

client/src/components/CreatePost.tsx
```diff
@@ -1,181 +1,279 @@
...
-    body: "",
-    file: "",
-    tags: "",
-    eventCapacity: "",
-    eventLocation: "",
+    title: '',
+    body: '',
+    file: '',
+    tags: '',
+    capacity: -1,
```
plausibly, line 26: 0

plausibly (changes requested):

Issue #38, opened 2021-11-09
Update models to keep track of user interactions with posts


plausibly:
We need to keep track of when users like/report/check-in to a specific post. This will help prevent spam

PR #64, opened 2021-11-26
Add ability to track which posts a user likes.

andy9775:
Create a new model which tracks when a user likes/unlikes a post. Additionally, add logic to like
and unlike posts. Finally, each `Post` response includes the number of likes and whether or not the
current user liked the given post.

This is part of #38

### DIFF
--- a/server/routes/v1/post.ts
+++ b/server/routes/v1/post.ts
@@ -4,7 +4,7 @@ import PostController from '../../controllers/v1/post';
 import { getAuthUser } from '../../middleware/auth';
 
 const postRouter = express.Router();
-const postController = new PostController(db.Post, db.userPostLikes);
+const postController = new PostController(db.Post, db.UserPostLikes);
 
 postRouter.get('', async (req: Request, res: Response) => {
   const limit = req.query.limit;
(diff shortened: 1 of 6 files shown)


andy9775:
Once the report functionality is in place, `feedbackScore` should either be a virtual field (calculated based on upvotes and reports) or it should be deprecated.

server/routes/v1/post.ts
```diff
@@ -18,9 +18,16 @@ postRouter.get('', async (req: Request, res: Response) => {
...
+      Number(limit),
+      Number(offset)
+    );
+    return res.status(result.status).json(result);
+  } catch (err) {
+    console.error(err);
+  }
 });
 
 postRouter.get('/:postid', async (req: Request, res: Response) => {
```
plausibly, line 33: getPost call should take in auth user id

server/routes/v1/post.ts
```diff
@@ -4,7 +4,7 @@ import PostController from '../../controllers/v1/post';
 import { getAuthUser } from '../../middleware/auth';
 
 const postRouter = express.Router();
-const postController = new PostController(db.Post);
+const postController = new PostController(db.Post, db.userPostLikes);
```
plausibly, line 7: U is capital `db.UserPostLikes`

plausibly (changes requested):

plausibly (approved):

danielduong8 (approved):

gurvirsasan (approved):

PR #65, opened 2021-11-30
Create password reset component, email confirmation container, and password reset container

danielduong8:
- Create password reset component
  - Update "Forgot password" in sign in component and AuthContainer TabPanel to redirect to this component

- Create email confirmation container
  - Add route to containers in App.tsx

- Create password reset container
  - Add route to containers in App.tsx

- Add api handlers for sending password reset email, confirming an email, and resetting a password

- Update tests for user controller

- Update send reset email handler
  - Return statuses and messages
  - Prevent sending password reset when email is not confirmed

- Remove redundant password confirmation check in password reset handler

- Set jest config to ignore tests in build folder

- Disable verbose Sequelize logging to make jest tests and Heroku server logs easier to read

- Fix naming for email URLs

- Replace incorrect usage of `req.query` with `req.body`

- Remove unused `code` properties in data returns

- Add check for missing `process.env.PAGE_URL`

- Update `noAuthPaths` for new password reset and email confirmation handlers

- Update confirmation and password reset URLs in email service

This PR drove me insane ngl. Made me think there was an error in my code somewhere when really it was either Sendgrid or Outlook being garbage and randomly letting emails through and randomly blocking them. Last thing I did to make it work was put the email in my "safe senders" list on Outlook and that seemed to fix it, but then again, it was able to work without it as well on certain occasions so it could just completely be random. Would've pushed this a long time ago.

Tested to make sure everything works on Heroku, but will add Jest unit tests soon. Should still be good to review though.

Images:
image
image
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,10 @@ Then in the root directory of the project run this:
 
 Windows:
 
+`rmdir "%cd%/client/node_modules/models"`
+
+`rmdir "%cd%/client/node_modules/@types/models"`
+
 `mklink /D "%cd%/client/node_modules/models" "%cd%/server/build/models"`
 
 `mklink /D "%cd%/client/node_modules/@types/models" "%cd%/server/build/types/models"`
--- a/client/src/App.tsx
+++ b/client/src/App.tsx
@@ -79,7 +79,7 @@ function App() {
         <Routes>
           <Route path='/' element={<AuthContainer />} />
           <Route path='/dashboard' element={<ProtectedRoute />} />
-          <Route path='/password-reset' element={<PassResetContainer />} /> {/* make sure this matches with conf email */}
+          <Route path='/api/v1/users/password-reset' element={<PassResetContainer />} /> {/* make sure this matches with conf email */}
         </Routes>
       </BrowserRouter>
     </ThemeProvider>
--- a/client/src/App.tsx
+++ b/client/src/App.tsx
@@ -79,7 +79,8 @@ function App() {
         <Routes>
           <Route path='/' element={<AuthContainer />} />
           <Route path='/dashboard' element={<ProtectedRoute />} />
-          <Route path='/api/v1/users/password-reset' element={<PassResetContainer />} /> {/* make sure this matches with conf email */}
+          <Route path='/password-reset' element={<PassResetContainer />} /> {/* make sure this matches with conf email */}
+          <Route path='/email-confirmation' element={<PassResetContainer />} /> {/* make sure this matches with conf email */}
         </Routes>
       </BrowserRouter>
     </ThemeProvider>
--- a/client/src/App.tsx
+++ b/client/src/App.tsx
@@ -83,10 +83,7 @@ function App() {
         <Routes>
           <Route path='/' element={<AuthContainer />} />
           <Route path='/dashboard' element={<ProtectedRoute />} />
-          <Route
-            path='/email-confirmation'
-            element={<EmailConfirmContainer />}
-          />
+          <Route path='/confirm-account' element={<EmailConfirmContainer />} />
           <Route path='/password-reset' element={<PassResetContainer />} />
         </Routes>
       </BrowserRouter>
--- a/client/src/App.tsx
+++ b/client/src/App.tsx
@@ -5,7 +5,7 @@ import CssBaseline from '@mui/material/CssBaseline';
 import { createTheme, ThemeProvider } from '@mui/material/styles';
 
 import PostDashboard from './containers/PostDashboard';
-import { AuthContainer } from './containers';
+import { AuthContainer, PassResetContainer } from './containers';
 import './App.css';
 
 import { User } from 'models/user';
@@ -79,6 +79,7 @@ function App() {
         <Routes>
           <Route path='/' element={<AuthContainer />} />
           <Route path='/dashboard' element={<ProtectedRoute />} />
+          <Route path='/password-reset' element={<PassResetContainer />} /> {/* make sure this matches with conf email */}
         </Routes>
       </BrowserRouter>
     </ThemeProvider>
--- a/client/src/App.tsx
+++ b/client/src/App.tsx
@@ -5,7 +5,7 @@ import CssBaseline from '@mui/material/CssBaseline';
 import { createTheme, ThemeProvider } from '@mui/material/styles';
 
 import PostDashboard from './containers/PostDashboard';
-import { AuthContainer, PassResetContainer } from './containers';
+import { AuthContainer, EmailConfirmContainer, PassResetContainer } from './containers';
 import './App.css';
 
 import { User } from 'models/user';
@@ -79,8 +79,8 @@ function App() {
         <Routes>
           <Route path='/' element={<AuthContainer />} />
           <Route path='/dashboard' element={<ProtectedRoute />} />
+          <Route path='/email-confirmation' element={<EmailConfirmContainer />} /> {/* make sure this matches with conf email */}
           <Route path='/password-reset' element={<PassResetContainer />} /> {/* make sure this matches with conf email */}
-          <Route path='/email-confirmation' element={<PassResetContainer />} /> {/* make sure this matches with conf email */}
         </Routes>
       </BrowserRouter>
     </ThemeProvider>
--- a/client/src/App.tsx
+++ b/client/src/App.tsx
@@ -5,7 +5,11 @@ import CssBaseline from '@mui/material/CssBaseline';
 import { createTheme, ThemeProvider } from '@mui/material/styles';
 
 import PostDashboard from './containers/PostDashboard';
-import { AuthContainer, EmailConfirmContainer, PassResetContainer } from './containers';
+import {
+  AuthContainer,
+  EmailConfirmContainer,
+  PassResetContainer,
+} from './containers';
 import './App.css';
 
 import { User } from 'models/user';
@@ -79,8 +83,11 @@ function App() {
         <Routes>
           <Route path='/' element={<AuthContainer />} />
           <Route path='/dashboard' element={<ProtectedRoute />} />
-          <Route path='/email-confirmation' element={<EmailConfirmContainer />} /> {/* make sure this matches with conf email */}
-          <Route path='/password-reset' element={<PassResetContainer />} /> {/* make sure this matches with conf email */}
+          <Route
+            path='/email-confirmation'
+            element={<EmailConfirmContainer />}
+          />
+          <Route path='/password-reset' element={<PassResetContainer />} />
         </Routes>
       </BrowserRouter>
     </ThemeProvider>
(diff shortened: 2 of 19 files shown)


client/src/App.tsx
```diff
@@ -79,6 +79,8 @@ function App() {
         <Routes>
           <Route path='/' element={<AuthContainer />} />
           <Route path='/dashboard' element={<ProtectedRoute />} />
+          <Route path='/email-confirmation' element={<EmailConfirmContainer />} /> {/* make sure this matches with conf email */}
```
andy9775, line 82:
```suggestion
          <Route path='/confirm-account' element={<EmailConfirmContainer />} /> {/* make sure this matches with conf email */}
```

client/src/api/v1/index.ts
```diff
@@ -82,11 +82,19 @@ export default class ServerApi {
     }
   }
 
-  protected async put<ResponseDataType>(
-    path: string
+  protected async put<ResponseDataType, RequestDataType=undefined>(
+    path: string, form?: RequestDataType
```
andy9775, line 86: https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads

client/src/api/v1/index.ts
```diff
@@ -155,22 +171,26 @@ export default class ServerApi {
...
+    >('/users/signin', form);
+  }
+
   async signOut() {
     return await this.post('/users/signout', {});
   }
 
-  async likePost(postID: string) {
-    return await this.put<{ status: number }>(`/posts/${postID}/upvote`);
+  async sendPassReset(form: { email: string }) {
```
andy9775, line 185:
```suggestion
  async requestPasswordReset(form: { email: string }) {
```

client/src/api/v1/index.ts
```diff
@@ -155,22 +171,26 @@ export default class ServerApi {
...
+  async confirmEmail(form: { token: string }) {
+    return await this.put<{ message: string }, typeof form>('/users/confirm-email', form);
   }
 
-  async signIn(form: { userName: string; password: string }) {
-    return await this.post<
-      typeof form,
-      { message: string; result?: UserAttributes }
-    >('/users/signin', form);
+  async resetPass(form: { password: string, token: string }) {
```
andy9775, line 193:
```suggestion
  async resetPassword(form: { password: string, token: string }) {
```

client/src/components/SignIn.tsx
```diff
@@ -5,8 +5,8 @@ import Avatar from '@mui/material/Avatar';
 import TextField from '@mui/material/TextField';
 import Button from '@mui/material/Button';
 import Box from '@mui/material/Box';
-import FormControlLabel from '@mui/material/FormControlLabel';
-import Checkbox from '@mui/material/Checkbox';
+// import FormControlLabel from "@mui/material/FormControlLabel";
+// import Checkbox from "@mui/material/Checkbox";
```
andy9775, line 9:
```suggestion
```

client/src/components/SignIn.tsx
```diff
@@ -193,10 +193,11 @@ function SignIn(props: { handleChange: Function; showAlert: Function }) {
...
         />
-        <FormControlLabel
-          control={<Checkbox value='remember' color='primary' />}
-          label='Remember me'
-        />
+        {/* will be implemented in future PR (after search) */}
+        {/* <FormControlLabel
+          control={<Checkbox value="remember" color="primary" />}
+          label="Remember me"
+        /> */}
```
andy9775, line 200:
```suggestion
```

client/src/components/PassReset.tsx
```diff
@@ -0,0 +1,187 @@
...
+          sx={{ mt: 3, mb: 2 }}
+        >
+          Reset my password
+        </Button>
+        <Grid container>
+          <Grid item>
+            <Link
+              href="#"
+              variant="body2"
+              onClick={(e) => props.handleChange(e, 0)}
```
andy9775, line 175: I'm assuming that the `0` here is the index? If so, this should be passed in, or handleChange should only take an error and the container which uses this component should worry about the index

server/server.ts
```diff
@@ -23,8 +29,9 @@ const noAuthPaths = [
   '/signin',
   '/signup',
   '/signout',
-  '/confirm',
-  '/password-reset',
+  '/send-password-reset',
```
andy9775, line 32:
```suggestion
  '/request-password-reset',
```

server/routes/v1/user.ts
```diff
@@ -91,44 +91,48 @@ export async function signUpHandler(req: Request, res: Response) {
...
-    res
-      .status(400)
-      .json({ code: 400, message: 'Token is invalid or expired.' });
+    res.status(400).json({ message: 'Token is invalid or expired.' });
   }
 }
 
 async function resetPassHandler(req: Request, res: Response) {
-  if (!req.query.r || !req.body.password || !req.body.password_confirmation) {
+  if (!req.body.token || !req.body.password) {
```
andy9775, line 120: Password reset and account confirmation should both take a password and password confirmation.

server/models/index.ts
```diff
@@ -12,7 +12,9 @@ if (
   !process.env.DATABASE_URL
 ) {
   /* For CI testing */
-  sequelize = new Sequelize('sqlite::memory');
+  sequelize = new Sequelize('sqlite::memory', {
+    logging: false,
```
andy9775, line 16:
```suggestion
```

It's useful to see what's going on.

server/controllers/v1/tests/user.test.ts
```diff
@@ -205,7 +205,9 @@ describe('v1 - User Controller', () => {
         fail('nullperson');
       }
 
-      expect(testPerson.confirmed).toBeFalsy();
+      testPerson.confirmed = true;
+      await testPerson.save();
+      expect(testPerson.confirmed).toBeTruthy();
```
andy9775, line 210: This isn't really adding any value. Your setting a field to true in the test and ten checking that its true.

server/controllers/v1/tests/user.test.ts
```diff
@@ -234,20 +236,9 @@ describe('v1 - User Controller', () => {
       await expect(uContr.confirmEmail(rawToken)).resolves.toBeFalsy();
     });
 
-    test('Ensure nothing happens if the new password is invalid', async () => {
```
andy9775, line 237: why remove?

andy9775 (changes requested):
Just a few comments after a first pass.

Also, make sure you're formatting correctly i.e. `PassReset.tsx` is using double and not single quotes, and I also noticed a mix of the two in another file. You can setup your vs code to format on save and it should automatically make the changes for you every time.

client/src/components/PassReset.tsx
```diff
@@ -0,0 +1,187 @@
...
+        <TextField
+          name="email"
+          label="Email Address"
+          onChange={handleFormInput}
+          fullWidth
+          required
+          data-testid="resetField"
+          type="email"
+          onBlur={(e) =>
+            validateBlur(
```
plausibly, line 152: Do we really need to validate this? We could have something like "No user exists with that email", or do what a lot of sites do and say "If an account is associated to that email, then a password reset has been sent" rather than client validators

client/src/containers/PassResetContainer.tsx
```diff
@@ -0,0 +1,251 @@
...
+            margin='normal'
+            required
+            fullWidth
+            name='password'
+            label='New Password'
+            type='password'
+            data-testid='password'
+            onChange={handleFormInput}
+            onBlur={(e) =>
+              validateBlur(
```
plausibly, line 216: add a RNG that has 1/100 chance to say "New password cannot be old password"

client/src/api/v1/index.ts
```diff
@@ -82,11 +82,19 @@ export default class ServerApi {
...
   }
 
-  protected async put<ResponseDataType>(
-    path: string
+  protected async put<ResponseDataType, RequestDataType=undefined>(
+    path: string, form?: RequestDataType
   ): Promise<{ status: number; data?: ResponseDataType }> {
     try {
-      const response = await this.api.put(path);
+      let response;
```
plausibly, line 89: Could we keep it consistent with how we did the previous functions and just set to empty object if there is no form?

server/routes/v1/user.ts
plausibly: account confirmation shouldn't require a password since all it does is confirming the email of an account with a given token

client/src/components/SignIn.tsx
danielduong8: I was worried about keeping this in, since, if we don't end up implementing this functionality, yet we have this button, we don't want them asking about it or to show them lol

danielduong8:
[#](https://github.com/GDSCUTM-CommunityProjects/UBoard/pull/65#discussion_r759884770)

This handler is not for account confirmation, it's for the password reset

client/src/containers/PassResetContainer.tsx
danielduong8: lol

client/src/components/PassReset.tsx
danielduong8: The validation is just for correct email format, the component already implements the "No user exists with that email" message.

server/controllers/v1/tests/user.test.ts
danielduong8: I removed this since this was a test for checking if a password matches the confirmation password. Since, I will have to add that confirmation validation back in, I will also add this test back in.

server/controllers/v1/tests/user.test.ts
danielduong8: The truthy thing was sort of a vestigial expect statement, I should have removed it tbh

server/models/index.ts
danielduong8: I was supposed to revert this commit after but I forgot to

server/routes/v1/user.ts
danielduong8: This is the handler for resetting a password not account confirmation.

If by account confirmation you mean email verification, I don't think that should take a password and password confirmation, since all you really do is click a "Confirm email address" button.

client/src/components/PassReset.tsx
danielduong8: This index is for the TabPanel, and it works just like how the SignIn and SignUp components do for switching between tabs.

client/src/api/v1/index.ts
danielduong8: interesting o.o

andy9775 (approved):
Minor comment, otherwise LGTM!

plausibly (approved):

README.md
```diff
@@ -19,6 +19,10 @@ Then in the root directory of the project run this:
 
 Windows:
 
+`rmdir "%cd%/client/node_modules/models"`
+
+`rmdir "%cd%/client/node_modules/@types/models"`
+
```
plausibly, line 25: This should be in a separate PR.. Anyway, I don't think we need this.

plausibly (approved):

PR #67, opened 2021-12-01
Implement Tags on frontend & backend

plausibly:
Post controller & post routes:
- Added tags to backend (tag returned when getting post/posts)
- Tags can be created alongside posts (max of 3)
- Adjustments to post/tags model
- Other adjustments to make tags work with posts controller
- Update constraint to limit tags length to 15 characters, they should be short and simple.

Frontend
- Create post form allows inputting tags (seperated by spaces), added some styling to this
- Create post preview properly shows the tags inputted
- Post preview and reading a post properly fetches and displays tags from backend
- Tags are removed by clicking "x" or backspace when creating a post in the form
- Fixed a bug on creatform with the alert not properly clearing
- Prevent adding duplicate tags
- Prevent adding tags that are >15 chars in length

UI:
image
image
image

Text input is disabled after 3 tags, renabled when they delete a tag.


server/controllers/v1/post.ts
```diff
@@ -314,9 +339,36 @@ export default class PostController {
...
     }
 
+    if (tags) {
+      // could not get adding (multiple) tags to work any other way
+      const tagObjs = await Promise.all(
+        tags.map(async (text) => {
+          return this.setupTag(text.trim());
+        })
+      );
+      await post.addTags(tagObjs);
```
andy9775, line 349: The sql would be something like:
```sql
INSERT INTO Tags (text) VALUES ('firstTag', 'secondTag', 'thirdTag');
```

Alternatively, https://sequelize.org/v5/manual/instances.html#working-in-bulk--creating--updating-and-destroying-multiple-rows-at-once-

server/controllers/v1/post.ts
```diff
@@ -314,9 +339,36 @@ export default class PostController {
...
+   */
+  async setupTag(text: string): Promise<Tag> {
+    // findOrCreate is broken when using sqlite, so I had to make this function.
+    // https://github.com/sequelize/sequelize/issues/5535
+    const data = await this.tagsRepo.findByPk(text);
+    return data
+      ? data
+      : await this.tagsRepo.create({
+          text,
+        });
```
andy9775, line 369: The issue seems to be related to a test environment only. If we had more time ideally we would update the test environment to use a non-in-memory db to fix it, rather than hack a solution around to get around the test config.

server/controllers/v1/tests/post.test.ts
```diff
@@ -71,6 +74,86 @@ describe('Test v1 - Post Controller', () => {
...
+        'This is a new post!',
+        'This is a new post!This is a new post!',
+        'location',
+        10,
+        tags
+      );
+
+      if (result.data.result) {
+        const postTags = await getPostTags(result.data.result.id);
+        expect(postTags?.map((t) => t.text).sort()).toEqual(tags.sort());
```
andy9775, line 92: There should be a `toEqualSameMembers` matcher or something. Sorting is ignored.

server/controllers/v1/tests/post.test.ts
```diff
@@ -71,6 +74,86 @@ describe('Test v1 - Post Controller', () => {
...
+      } else {
+        expect(true).toBe(false);
```
andy9775, line 94: if statements in tests aren't good practice. You should expect that `result.data.result` is not empty/null in place of the `if(...)`

server/controllers/v1/tests/post.test.ts
```diff
@@ -71,6 +74,86 @@ describe('Test v1 - Post Controller', () => {
...
+      }
+
+      expect((await getAllTags()).length).toBe(tags.length);
+    });
+
+    it('getting a post will include tags', async () => {
```
andy9775, line 132: will include the associated tags

server/controllers/v1/tests/post.test.ts
```diff
@@ -71,6 +74,86 @@ describe('Test v1 - Post Controller', () => {
...
+        const postTags = await getPostTags(result.data.result.id);
+        expect(postTags?.map((t) => t.text).sort()).toEqual(tags.sort());
+      } else {
+        expect(true).toBe(false);
+      }
+
+      expect(result.status).toBe(200);
+    });
+
+    it('should be able to use existing tags', async () => {
```
andy9775, line 100: What's this testing? That tags aren't re-created? If so, that's more of a model test.

server/controllers/v1/tests/post.test.ts
```diff
@@ -128,6 +212,45 @@ describe('Test v1 - Post Controller', () => {
       expect(findResult.status).toBe(404);
     });
 
+    it('deleting a post maintains the same tags on another post', async () => {
```
andy9775, line 215: deleting a post doesn't delete tags.

server/routes/v1/post.ts
```diff
@@ -90,7 +90,8 @@ postRouter.post('/', async (req: Request, res: Response) => {
       req.body.title,
       req.body.body,
       req.body.location,
-      req.body.capacity
+      req.body.capacity,
+      req.body.tags.slice(0, 3) // max of 3 tags allowed
```
andy9775, line 94: this logic should exist in the controller - just pass in the tags here.

andy9775 (changes requested):
Good work!

One more comment, do we want to limit which users can create new tags? Future feature?

client/src/components/CreatePost.tsx
```diff
@@ -190,14 +201,64 @@ function CreatePost() {
...
+                          setForm({
+                            ...form,
+                            tags: [...form.tags, e.target.value.trim()],
+                          });
+                        } // new tag input for each space
+                        e.target.value = '';
+                      }
+                      setTagInputValue(e.target.value);
+                    }}
+                    onKeyDown={(e) => { // TODO bug
```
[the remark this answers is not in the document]
plausibly: Whoops, forgot to remove the TODO. I fixed the bug already

server/controllers/v1/tests/post.test.ts
plausibly: I know but I was getting TS errors if I didnt wrap it in this if statement

client/src/api/v1/index.ts
```diff
@@ -3,9 +3,15 @@ import axios, { AxiosInstance } from 'axios';
...
 import { User } from 'models/user';
 import { UserAttributes } from 'models/user';
+import { PostTag } from 'models/PostTags';
 
 export type PostUser = Post & {
+  likeCount: number;
+  doesUserLike: boolean;
   User: { id: string; firstName: string; lastName: string };
+  Tags: {
+    text: string & { PostTags: PostTag }; // sequelize pluarlizes name
```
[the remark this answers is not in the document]
plausibly: No, it was weird. I was doing manual testing and it returns a Tags object with text, and then a PostTags object. Maybe because I put `attributes: "text"` in my query? I'll look into the query to see if it can be cleaned up.

server/controllers/v1/post.ts
plausibly: I looked at bulk create, but is there a way to send an array through it? It seems like you have to actually pass the specific items being created. I could use .map() with the raw insert query, but I don't think it'll work with bulkCreate().  We aren't actually inserting the text into tags here, that's done in setupTag. Did you want me to change the query in setupTag()?

server/controllers/v1/post.ts
plausibly: Yeah.. if only :-(

client/src/components/CreatePost.test.tsx
```diff
@@ -63,4 +63,80 @@ describe('verifying launch of create post component', () => {
   it('closes the dialog when clicked on `Back` button', () => {
     expect(screen.getByTestId('newPostButton')).toBeInTheDocument();
   });
+
+  it('entering a tag and hitting space will add tag styling and empty the field', () => {
```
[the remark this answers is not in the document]
plausibly: No, we don't do anything with onBlur. I can adjust it so that whatever they typed into the box before clicking elsewhere will add the tag.

server/controllers/v1/tests/post.test.ts
plausibly: Just tried it and that didn't work... there doesn't seem to be a checker to ignore ordering, I checked jest docs and tried a couple other functions and none of them worked. Stackoverflow says to just use sort()

server/controllers/v1/tests/post.test.ts
plausibly: On second thought, I'll update it and just do @ts-ignore

client/src/api/v1/index.ts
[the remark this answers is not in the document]
andy9775: We define attributes for the other queries and they are fine.

server/controllers/v1/tests/post.test.ts
andy9775: Or if you know for sure that you'll get a result you can use `!`

client/src/components/CreatePost.test.tsx
```diff
@@ -63,4 +63,80 @@ describe('verifying launch of create post component', () => {
...
+    fireEvent.change(tagTextField, {
+      target: { value: 'tag2 ' },
+    });
+    fireEvent.change(tagTextField, {
+      target: { value: 'tag3 ' },
+    });
+    expect(tagTextField.disabled).toBeTruthy();
+  });
+
+  it('pressing backspace will delete the last tag', () => {
```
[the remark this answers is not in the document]
plausibly: I changed it so pressing backspace will put the tag back in the input box, incase they want to edit it

server/controllers/v1/tests/post.test.ts
plausibly: It's specifically testing if the controller is properly handling duplicaet tags

client/src/components/CreatePost.test.tsx
```diff
@@ -111,16 +111,13 @@ describe('verifying launch of create post component', () => {
     expect(tagTextField.disabled).toBeTruthy();
     expect(screen.getByText('tag1')).toBeInTheDocument();
     expect(screen.getByText('tag2')).toBeInTheDocument();
-    expect(screen.getByText('tag3')).toBeInTheDocument();
-
-    fireEvent.keyDown(tagTextField, {
-      key: 'Backspace',
-    });
+    expect(screen.getByText('tag2')).toBeInTheDocument();
+    fireEvent.click(screen.getByTestId('tag2-clicker'));
```
andy9775, line 115: tag2-clicker?

server/controllers/v1/post.ts
```diff
@@ -340,35 +340,22 @@ export default class PostController {
...
+        tags.slice(0, 3).map((t) => {
+          // restrict to max 3 tags
+          return { text: t.trim() };
+        }),
+        {
+          ignoreDuplicates: true,
+        }
       );
-      await post.addTags(tagObjs);
+      await post.addTags(tagObjs); // junction table association
```
andy9775, line 353:
```suggestion
      await post.addTags(tagObjs);
```
Comments should usually explain why you did something the way you did it, not what's happening.

server/controllers/v1/tests/post.test.ts
andy9775: It's not the controller that handles duplicates though. I would move it to the tag model test

server/controllers/v1/tests/post.test.ts
```diff
@@ -87,13 +87,42 @@ describe('Test v1 - Post Controller', () => {
...
+      const tags = ['csc108'];
+
+      await postController.createPost(
+        author.id,
+        'This is a new post!',
+        'This is a new post!This is a new post!',
+        'location',
+        10,
+        tags
+      );
```
andy9775, line 107: Should be able to create a tag without creating a post no?

server/controllers/v1/tests/post.test.ts
```diff
@@ -87,13 +87,42 @@ describe('Test v1 - Post Controller', () => {
...
+
+      expect((await getAllTags()).length).toBe(tags.length);
```
andy9775, line 109:
```suggestion
```

andy9775 (changes requested):
Did you write the onBlur test? Cause right now, if a user types a tag and clicks out that tag won't be applied which isn't a nice UX experience.

client/src/components/CreatePost.test.tsx
plausibly: The tags have an 'x' button on them which you can click which deletes the tag. Backspacing no longer deletes and puts the tag back in the input box instead.

The testid is for the x button inside Tag.tsx

server/controllers/v1/tests/post.test.ts
plausibly: No? We only create a tag if someone creates a post and puts that tag in it

plausibly:
> Did you write the onBlur test? Cause right now, if a user types a tag and clicks out that tag won't be applied which isn't a nice UX experience.

Yes, clicking away will apply the tag. There is a test 'entering a tag and unfocusing the input box will add a new tag' (line 81)

server/controllers/v1/tests/post.test.ts
plausibly: alright

server/controllers/v1/tests/post.test.ts
andy9775: I just meant you can create a tag via the model and skip the controller. Simplifies the test.

andy9775:
What am I reviewing? The last commit was 7 days ago?

plausibly:
> What am I reviewing? The last commit was 7 days ago?

No, I committed yesterday, is it not showing up?
image

server/controllers/v1/tests/post.test.ts
plausibly: I guess, but I also want to make sure that the tags can properly be associated to multiple posts in the junction table. Createpost will make the tag and that association for us.

andy9775 (approved):
LGTM!

PR #68, opened 2021-12-02
Implement search functionality for dashboard

danielduong8:
- Create client api handler for fetching via search
- Implement search
  - Send user input in search bar upon key press (Enter) or after a 1s delay
  - Call normal fetch handler if nothing is in search bar, otherwise call fetch handler with search query
  - Escape queries in frontend before passing to backend
  - Use raw queries and temporary tables to implement search in backend 
    - Assign weights from A-D (highest-lowest) to title, location, and body, respectively
      - Weights are scaled by document length, number of unique words, and the mean harmonic distance between extents
    - Search database based on input query and rank the results based on relevance
    - Ranking is determined by cover density (proximity of matching lexemes) and weights
  - Change routing logic for getting posts to be dependent on whether a query is passed
- Increase polling interval from 0.5s to 5s
- Add double quotes to literal query to resolve db errors (pg requires double quotes for columns or tables with uppercase letters) (will also put this in a separate PR as well)

Same as with my other PR, tested and fully functional on Heroku, but will add jest unit tests later.


client/src/api/v1/index.ts
```diff
@@ -107,6 +107,20 @@ export default class ServerApi {
     return await this.get<{}, User>('/users/me', {});
   }
 
+  async fetchPostsBySearch(query: string, limit: number, offset: number) {
```
andy9775, line 110: `findPosts` `searchForPosts`?

server/controllers/v1/post.ts
```diff
@@ -61,15 +62,15 @@ export default class PostController {
           'thumbnail',
           [
             sequelize.literal(
-              `(SELECT COUNT(*) FROM UserPostLikes as Likes WHERE Likes.postID = Post.id)`
+              `(SELECT COUNT(*) FROM "UserPostLikes" as "Likes" WHERE "Likes"."postID" = "Post"."id")`
```
andy9775, line 65: Necessary? Is this a heroku issue?

server/controllers/v1/post.ts
```diff
@@ -105,6 +106,74 @@ export default class PostController {
...
+    await db.sequelize.query(`DROP TABLE IF EXISTS "PostSearches";`);
+    await db.sequelize.query(
+      `CREATE TEMP TABLE "PostSearches" AS 
+      SELECT *, (
+        setweight(to_tsvector(coalesce("title",'')), 'A') || 
+        setweight(to_tsvector(coalesce("location",'')), 'C') || 
+        setweight(to_tsvector(coalesce("body",'')), 'D')
+      ) as "tsvector" 
+      FROM "Posts";`
+    );
```
andy9775, line 127: We don't want to do this on each and every request. 

I'm assuming that the PostSearches table is used to cache the search query. This means we have to build the table per query but what if we have lots of data? How long will that take? 

Second, what if I make a query, the `PostSearches` table is created and built, then before the `select` is executed below for me, you make a search and `DROP TABLE IF EXISTS "PostSearches"` is run and then my `select` is run?

server/controllers/v1/post.ts
```diff
@@ -105,6 +106,74 @@ export default class PostController {
...
+    return {
+      status: data[0].length > 0 ? 200 : 204,
+      data: {
+        result: data[0].map((p) => {
+          // Must case to `any` as dataValues is not typed at the moment.
+          // Context: https://github.com/RobinBuschmann/sequelize-typescript/issues/760
+          // p.likeCount = (p as any).dataValues.likeCount;
+          // p.doesUserLike = (p as any).dataValues.doesUserLike == 1;
+          return p;
+        }),
```
andy9775, line 170: no-op map call???

server/routes/v1/post.ts
```diff
@@ -7,6 +7,7 @@ const postRouter = express.Router();
 const postController = new PostController(db.Post, db.UserPostLikes);
 
 postRouter.get('', async (req: Request, res: Response) => {
```
andy9775, line 9: create a `/search` endpoint.

server/controllers/v1/post.ts
```diff
@@ -105,6 +106,74 @@ export default class PostController {
     };
   }
 
+  async getPostsByQuery(
```
andy9775, line 109:
```suggestion
  async searchForPosts(
```

andy9775 (changes requested):
Just some comments after a first pass

server/routes/v1/post.ts
danielduong8: That was actually what I did first, but it ended up conflicting with the /posts/:postid/ endpoints when you actually call something like uboard.com/posts/search

server/controllers/v1/post.ts
danielduong8: Wasn't sure if it was needed or not, but I guess now I know lol

Also looks like those properties are still present in the response when looking at the dev console

server/controllers/v1/post.ts
danielduong8: Postgres issue

client/src/containers/PostDashboard.tsx
```diff
@@ -41,16 +51,28 @@ function RecentPosts(props: {
           })
           .catch((err) => console.log(err));
       }
-    }, 500);
+    }, 5000);
```
plausibly, line 54: This should be a separate PR? Additionally, if you increase polling then we should add something to instantly fetch posts when you create a post. Otherwise it will look like there is a 5 second lag when a post is made

plausibly (changes requested):

client/src/containers/PostDashboard.tsx
danielduong8: I think you already have that implemented, isn't that why you have 2 useEffects()?

server/routes/v1/post.ts
danielduong8: Nvm figured it out, just had to move the route to the top of the file (ahove the :postid routes)

server/controllers/v1/post.ts
```diff
@@ -115,38 +115,51 @@ export default class PostController {
...
+    // await db.sequelize.query(`DROP TABLE IF EXISTS "PostSearches";`);
+    // await db.sequelize.query(
+    //   `CREATE TEMP TABLE "PostSearches" AS
+    //   SELECT *, (
+    //     setweight(to_tsvector(coalesce("title",'')), 'A') ||
+    //     setweight(to_tsvector(coalesce("location",'')), 'C') ||
+    //     setweight(to_tsvector(coalesce("body",'')), 'D')
+    //   ) as "tsvector"
+    //   FROM "Posts";`
+    // );
```
andy9775, line 127: Remove unused code

server/controllers/v1/post.ts
```diff
@@ -115,38 +115,51 @@ export default class PostController {
...
+          "Post"."body", 
+          "Post"."title", 
+          "Post"."createdAt", 
+          "Post"."thumbnail", 
           (SELECT COUNT(*) FROM "UserPostLikes" AS "Likes" 
-            WHERE "Likes"."postID" = "PostSearch"."id") AS "likeCount", 
+            WHERE "Likes"."postID" = "Post"."id") AS "likeCount", 
           (SELECT COUNT(*) FROM "UserPostLikes" AS "Likes" 
-            WHERE "Likes"."postID" = "PostSearch"."id" AND "Likes"."userID" = $1) AS "doesUserLike", 
+            WHERE "Likes"."postID" = "Post"."id" AND "Likes"."userID" = $1) AS "doesUserLike",
```
andy9775, line 139: instead of $1 I would used named variables in such a large query e.g. `:userid`

andy9775 (changes requested):
I would add some tests given the custom query. Create a `search.test.ts` file, create a bunch of fake posts in the setup function, then have a couple (3-4) tests which search for some terms. It could be a term which appears throughout a text (e.g. `add tests query` would be good one for this comment), a term which appears next to another, maybe a test for ranking.

server/controllers/v1/post.ts
```diff
@@ -105,6 +106,81 @@ export default class PostController {
...
+            WHERE "Likes"."postID" = "Post"."id") AS "likeCount", 
+          (SELECT COUNT(*) FROM "UserPostLikes" AS "Likes" 
+            WHERE "Likes"."postID" = "Post"."id" AND "Likes"."userID" = $1) AS "doesUserLike", 
+          json_build_object(
+            'firstName', "User"."firstName", 
+            'lastName', "User"."lastName",
+            'id', "User"."id"
+          ) AS "User",
+          (
+            setweight(to_tsvector(coalesce("Post"."title",'')), 'A') ||
```
plausibly, line 146: Can we also search by an author? I.e username or firstname lastname? If you implement that, can you update `My Posts` (on the menu bar) to search by the current user

server/controllers/v1/post.ts
```diff
@@ -105,6 +106,81 @@ export default class PostController {
...
+          ) AS "tsvector",
+          ts_rank_cd((
+            setweight(to_tsvector(coalesce("Post"."title",'')), 'A') || 
+            setweight(to_tsvector(coalesce("Post"."location",'')), 'C') || 
+            setweight(to_tsvector(coalesce("Post"."body",'')), 'D')
+          ), "query", 1|4) AS "rank" 
+        FROM "Posts" AS "Post" 
+        LEFT OUTER JOIN "Users" AS "User" ON "Post"."UserId" = "User"."id" 
+        CROSS JOIN to_tsquery($2) AS "query" 
+        WHERE "query" @@ (
```
plausibly, line 158: Is `"query"` the user input sent from the client? Shouldn't we sanitize this since raw sequelize queries don't have SQLI protection?

client/src/containers/PostDashboard.tsx
```diff
@@ -78,9 +100,21 @@ export default function PostDashboard() {
   const [pageCount, setPageCount] = React.useState(1);
   const [page, setPage] = React.useState(1);
 
+  const useQuery = (): [string, (q: string) => void] => {
+    const [query, setQuery] = React.useState('');
+
+    const setEscapedQuery = (q: string) => {
+      setQuery(q.replace(/\W+/g, '&'));
```
plausibly, line 107: Is this the only place we sanatize user input? Shouldn't we also do this in the backend, since nothing is stopping the user from directly calling our route

plausibly (changes requested):

server/controllers/v1/post.ts
danielduong8: prepared statements should be enough to circumvent sqli

client/src/components/Header.tsx
```diff
@@ -95,7 +98,27 @@ function AccountMenu() {
           'aria-labelledby': 'menu-btn',
         }}
       >
-        <MenuItem onClick={closeMenu}>My Posts</MenuItem>
+        <MenuItem
+          onClick={async () => {
```
plausibly, line 102: instead of async can you do .then? we did that for most of our api calls on f/e

client/src/components/Header.tsx
```diff
@@ -95,7 +98,27 @@ function AccountMenu() {
...
+            try {
+              const result = await api.me();
```
andy9775, line 104: Doesn't the current user info exist in context to be shared across the application? We shouldn't need to hit `me` again.

@plausibly

client/src/components/Header.tsx
```diff
@@ -95,7 +98,27 @@ function AccountMenu() {
...
+              const user = result.data;
+              if (!user) {
+                throw new Error('Current user could not be found.');
+              }
+
+              const query = `${user.firstName} ${user.lastName}`;
+              props.setSearch(query);
```
andy9775, line 111: What happens if a user is mentioned in a post or comment? Does that return for `My Posts`? IMO there should be a myPosts endpoint on the API for this functionality.

andy9775 (approved):
Just a few comments.

I'm not sure about the functionality of the search so I think we should write some tests for it. You can create a bunch of fake posts (realistic ones) and then write some tests to ensure that the correct ones are returned for a specific query. This shouldn't really take more than an hour, just create ~15 unique posts then have ~5 tests covering things like ordering (ensure that the relevant post is at the top), addball queries, things like that.

client/src/components/Header.tsx
plausibly: Oh yea, that is true. On `App.tsx` there is `UserContext`

danielduong8:
> Just a few comments.
> 
> I'm not sure about the functionality of the search so I think we should write some tests for it. You can create a bunch of fake posts (realistic ones) and then write some tests to ensure that the correct ones are returned for a specific query. This shouldn't really take more than an hour, just create ~15 unique posts then have ~5 tests covering things like ordering (ensure that the relevant post is at the top), addball queries, things like that.

Sadly, I cannot test the search this way since my query uses psql-specific functions and jest does tests with sqlite. I have tested it quite thoroughly on the actual app though and it seems to be working fine.

plausibly (approved):

PR #70, opened 2021-12-02
Create CRUD endpoints for comments.

gurvirsasan:
Create the CRUD controller to:

- getComments
- getComment (by ID)
- Delete comment
- update comment
- create comment
 
connects up the controller methods to the routes.


server/routes/v1/index.ts
```diff
@@ -1,10 +1,12 @@
 import express from 'express';
 import userRouter from './user';
 import postRouter from './post';
+import commentsRouter from './comment';
 
 const routes = express.Router();
 
 routes.use('/users/', userRouter);
 routes.use('/posts/', postRouter);
+routes.use('/comments/', commentsRouter);
```
andy9775, line 10:
```suggestion
routes.use('/posts/:postid/comments/', commentsRouter);
```

You can now access the post the comment is being created for by `postid`

server/routes/v1/comment.ts
```diff
@@ -0,0 +1,85 @@
...
+      req.body.body
+    );
+    res.status(result.status).json(result);
+  } catch (err) {
+    console.error(err);
+  }
+});
+
+commentRouter.put(
+  '/postid/comments/:commentid',
```
andy9775, line 69:
```suggestion
  '/:commentid',
```

server/routes/v1/comment.ts
```diff
@@ -0,0 +1,85 @@
...
+      getAuthUser(res).id,
+      req.params.commentid
+    );
+    res.status(result.status).json(result);
+  } catch (err) {
+    console.error(err);
+  }
+});
+
+commentRouter.post(':postid/comments/', async (req: Request, res: Response) => {
```
andy9775, line 55:
```suggestion
commentRouter.post('', async (req: Request, res: Response) => {
```

server/controllers/v1/comment.ts
```diff
@@ -0,0 +1,211 @@
+import { Comment } from '../../models/comment';
+import db from '../../models';
+
+// The return type of a Comment associated with the Post's User.
+export type CommentsUser = Comment & {
+  User: {
+    id: string;
+    firstName: string;
+    lastName: string;
+    createdAt: string;
```
andy9775, line 10: No need for `createdAt` for the user data.
Here and elsewhere.

server/controllers/v1/tests/comment.test.ts
```diff
@@ -0,0 +1,160 @@
...
+        author.id,
+        comment.id
+      );
+      expect(result.status).toBe(204);
+
+      const findResult = await commentController.getComment(comment.id);
+      expect(findResult.status).toBe(404);
+    });
+
+    it('should not delete a different comment', async () => {
```
andy9775, line 118: `should not delete a comment not authored by a user`?

server/controllers/v1/tests/comment.test.ts
```diff
@@ -0,0 +1,160 @@
...
+        author.id,
+        comment.id,
+        newBody
+      );
+
+      expect(result.status).toBe(200);
+      expect(result.data!.result!.body).toBe(newBody);
+    });
+
+    it('should not update the comment for invalid user', async () => {
```
andy9775, line 86: `should only update a users comment`?

server/controllers/v1/tests/comment.test.ts
```diff
@@ -0,0 +1,160 @@
...
+      const result = await commentController.createComment(
+        'This is a new comment!This is a new comment!!',
+        author.id,
+        undefined
+      );
+      expect(result.status).toBe(400);
+    });
+  });
+
+  describe('post mutation', () => {
```
andy9775, line 69: `comment mutation`

server/controllers/v1/tests/comment.test.ts
```diff
@@ -0,0 +1,160 @@
...
+        expect(result.data.result![index].id).toBe(comments[index].id);
+        expect(result.data.result![index].User.firstName).toBe(
+          author.firstName
+        );
+      }
+      expect(result.data.total).toBe(3);
+    });
+
+    it('should return an error for a missing comment', async () => {
+      const result = await commentController.getComment('bingBong');
```
andy9775, line 40:
```suggestion
      const result = await commentController.getComment('123-123');
```

This follows the structure of a UID making it more obvious what it is

andy9775 (changes requested):
Good job, some minor comments.

server/controllers/v1/tests/comment.test.ts
```diff
@@ -25,19 +25,19 @@ describe('Test v1 - Comment Controller', () => {
...
-        expect(result.data.result![index].User.firstName).toBe(
-          author.firstName
-        );
-      }
+      expect(result.data.result!.map((x) => x.id).sort()).toEqual(
+        comments.map((x) => x.id).sort()
+      );
+      expect(
+        result.data.result!.map((x) => x.User.firstName === author.firstName)
+      ).toBeTruthy();
```
andy9775, line 34:
```suggestion
      expect(result.data.result!.map((c) => c.id).sort()).toEqual(
        comments.map((c) => c.id).sort()
      );
      expect(
        result.data.result!.map((c) => c.User.firstName === author.firstName)
      ).toBeTruthy();
```
`c` for comment, we're not in math lol

server/controllers/v1/tests/comment.test.ts
```diff
@@ -83,7 +83,7 @@ describe('Test v1 - Comment Controller', () => {
       expect(result.data!.result!.body).toBe(newBody);
     });
 
-    it('should not update the comment for invalid user', async () => {
+    it("should only update the author's comment", async () => {
```
andy9775, line 86:
```suggestion
    it('should only update the author's comment', async () => {
```

Enable format on save in vs code :)

andy9775 (approved):
Some minor nits, otherwise LGTM!

server/controllers/v1/tests/comment.test.ts
gurvirsasan: I have that enabled, but with `\'s` it automatically results to the double quotes

server/routes/v1/comment.ts
```diff
@@ -0,0 +1,82 @@
...
+import db from '../../models';
+import CommentController from '../../controllers/v1/comment';
+import { getAuthUser } from '../../middleware/auth';
+
+const commentRouter = express.Router();
+const commentController = new CommentController(db.Post);
+
+commentRouter.get('', async (req: Request, res: Response) => {
+  try {
+    const limit = req.body.limit;
```
plausibly, line 11: These should be `req.query` for a GET request, we cannot send a body.

server/routes/v1/comment.ts
```diff
@@ -0,0 +1,82 @@
...
+  } catch (err) {
+    console.error(err);
+  }
+});
+
+commentRouter.post('', async (req: Request, res: Response) => {
+  try {
+    const result = await commentController.createComment(
+      getAuthUser(res).id,
+      req.params.postid,
```
plausibly, line 59: shouldnt this be req.body?

plausibly (changes requested):

server/controllers/v1/tests/comment.test.ts
andy9775: NVM. I didn't realize that there was an apostrophe. Ignore me

server/routes/v1/comment.ts
gurvirsasan: makes sense!

server/controllers/v1/comment.ts
```diff
@@ -0,0 +1,218 @@
+import { Comment } from '../../models/comment';
+import db from '../../models';
+
+// The return type of a Comment associated with the Post's User.
+export type CommentsUser = Comment & {
+  User: {
```
plausibly, line 6: Check in postman to make sure this type is correct. I remember we had a similar mistake for PostUser, it was returning `UserId` separately from User

server/controllers/v1/comment.ts
```diff
@@ -0,0 +1,218 @@
...
+      this.commentsRepo.count({ where: { PostId: postID } }),
+    ]);
+
+    if (!data[0]) {
+      return {
+        status: 404,
+        data: { message: `Post ${postID} could not be found` },
+      };
+    }
+    if (!data[1]) {
```
plausibly, line 65: Why not just combine this if statement with the same one on line 59 with an `or`? We are returning `404` for both, so just update the status message with a general error message

plausibly (approved):
LGTM. Just some minor comments

plausibly:
Note: need to update getPosts to return number of comments in that post

server/controllers/v1/comment.ts
```diff
@@ -11,7 +11,7 @@ export type CommentsUser = Comment & {
 };
 
 // The maximum number of results to return.
-const MAX_RESULTS = 50;
+const MAX_RESULTS = 10;
```
andy9775, line 14: Only 10 posts per query?

andy9775 (approved):
One minor comment: Normally you wouldn't/shouldn't change existing migration files. You would create a new migration. Since we're dumping the DB each time, it's fine. Just an FYI.

Otherwise LGTM!

server/controllers/v1/comment.ts
plausibly: Yea, I think this is better than having lots of scrolling on the post page

server/models/tests/testHelpers.ts
```diff
@@ -153,13 +152,24 @@ export async function makeValidComment(
...
+export async function getAllTags(): Promise<Tag[]> {
+  return await TagModel.findAll();
+}
+
+export async function getPostTags(
+  postId: string
+): Promise<(Tag & PostTag)[] | undefined> {
+  const post = await PostModel.findByPk(postId);
+
+  return await post?.getTags();
```
danielduong8, line 171: couldn't post be null if no postId is found? then you would call getTags() on null

danielduong8 (approved):
lgtm

server/models/tests/testHelpers.ts
plausibly: its just for tests, but this looks like it was from the Tags PR.

PR #72, opened 2021-12-16
 Front-End Google Maps Integration

plausibly:
Note: Need to update the ENV variable, other than that the PR is ready for review

Works in conjunction with PR #76 

- Integrated Google Maps API on the frontend (Geocoding, Places API, Maps for JS)
- Hoping to god that we will not run out of 'free' queries on Google Cloud

Create Post:
- Added auto textbox which suggestions locations based on addresses, or names of places (I.e University of Toronto)
- (collapsible) Interactive map with a draggable pin, allows setting a precise location. This change is reflected in the autofill box
- Ability to select online event, this will disable the maps features

View Post:
- (collapsible) Added map showing location of the event, marked by a pin
- Map can be zoomed into, full screen, street view'd, etc..
- The actual location (i.e address) is shown in text above the map
- For online events, the map features are not shown.

The backend will be done in a seperate PR (#76 ). This requires changing the Post model and the controllers. I will need to add
coordinates to our model to allow for storing coordinates `{ lat, long }` so that it can be used to load the map. The existing `location` attribute will store the address of the location, or the link to the online event.

Note: For events that are online, lat and long are set to -1 to indicate not to use google maps.

Create Post:
image

Autocomplete:
image
image

Interactive map (Move around pin to update location textbox):
image

Hide Map:
image

Online Event (Disable maps features):
image

Viewing a post:

image
image
(For online events, the switch won't show up -- it will just say Location)


client/src/components/CreatePost.tsx
```diff
@@ -33,8 +42,37 @@ function CreatePost() {
     tags: [] as string[],
     capacity: 0,
     location: '',
+    coords: { lat: -1, lng: -1 }, // lat/lng -1 indicates online event.
   });
 
+  const handleLocation = (
```
andy9775, line 48:
```suggestion
  const updateLocation = (
```

client/src/components/CreatePost.tsx
```diff
@@ -33,8 +42,37 @@ function CreatePost() {
...
+    setForm((form) => {
+      return {
+        ...form,
+        location: location.location,
+        coords: { lat: location.lat, lng: location.lng },
+      };
+    });
+  }, [location]);
+
+  const handleLocationToggle = (e: React.ChangeEvent<HTMLInputElement>) => {
```
andy9775, line 71:
```suggestion
  const toggleOnlineLocation = (e: React.ChangeEvent<HTMLInputElement>) => {
```

client/src/components/ViewPostDialog.tsx
```diff
@@ -206,6 +208,39 @@ function CapacityBar(props: { maxCapacity: number }) {
   );
 }
 
+function LocationHandler(props: {
```
andy9775, line 211:
```suggestion
function LocationMap(props: {
```

client/src/components/ViewPostDialog.tsx
```diff
@@ -206,6 +208,39 @@ function CapacityBar(props: { maxCapacity: number }) {
...
+  coords?: { lat: number; lng: number };
+  location: string;
+}) {
+  const [isMapVisible, toggleMap] = React.useState(true);
+  const mapEnabled =
```
andy9775, line 216:
```suggestion
  const isOfflineEvent =
```

andy9775 (changes requested):
Looks good. 

I would also add a test for toggling the map and other user-map interactions you can think of.

client/src/components/ViewPostDialog.tsx
plausibly: LocationMap is already taken, it's the component imported from `LocationMap.tsx`

danielduong8 (approved):

client/src/components/LocationMap.tsx
```diff
@@ -90,16 +110,18 @@ export function LocationPickerMap( props: { setLocation:  (location: string, lat
       </FormGroup>
       <Paper
         elevation={5}
-        style={{ height: showMap ? '35vh' : '0', width: '100%' }}
+        style={{ height: showMap ? '35vh' : '0', width: '100%' }} // don't unload the map, 0 height will hide it
```
andy9775, line 113: usually we want to unload and not hide, save some resources.

andy9775 (approved):
Just on minor nit, otherwise LGTM!

client/src/components/LocationMap.tsx
plausibly: Can't unload it, the component loads the google maps API, unloading it also unloads the API.

PR #73, opened 2021-12-18
Style adjustment to tags

plausibly:
Slightly adjusted tag creation to make it more user friendly.

- Hitting 'Enter' creates tags now
- Adjusted tag deletion so 'X' does not put it back in the input box, but backspace does.


client/src/components/CreatePost.tsx
```diff
@@ -35,9 +35,12 @@ function CreatePost() {
     location: '',
   });
 
-  const handleTagDelete = (value: string) => {
+  const handleTagDelete = (value: string, emptyInput: boolean = false) => {
```
andy9775, line 38:
```suggestion
  const handleTagDelete = (value: string, clearInput: boolean = false) => {
```

andy9775 (approved):
Minor nit, otherwise LGTM!

PR #76, opened 2021-12-21
Update backend to implement map coordinates

plausibly:
- Post model has `coords` with type JSON in the format `{lat: number; lng: number}`
- By default coords will be `{lat: -1; lng: -1}`, we can also check this in the frontend. -1 will indicate an online event
- getPost will return the `coords` of a post
- Updated create and update post to handle coords
- Routes have been updated to implement coords

We will be using the `location` to store the exact address of an event in string format (or the location of the online event) and is shown to the user when viewing a post. Coords will store the latitude and longitude and won't be explicitly shown to the user, but instead used to fetch the location with Google Maps API.

### DIFF
--- a/server/models/tests/post.test.ts
+++ b/server/models/tests/post.test.ts
@@ -18,11 +18,13 @@ describe('Post Model', () => {
       expect(author).toBeDefined();
 
       expect(author.userName).toContain('postTester');
+      const coord = { lat: 34.123913028321, lng: -96.123913028321 };
 
       const posted = await makePost(
         author.id,
         'This is a new post!',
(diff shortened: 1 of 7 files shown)


server/models/tests/post.test.ts
```diff
@@ -18,18 +18,21 @@ describe('Post Model', () => {
       expect(author).toBeDefined();
 
       expect(author.userName).toContain('postTester');
+      const coord = { lat: 34.123913028321, lng: -96.123913028321 };
```
andy9775, line 21: Additional test for an online event?

andy9775 (approved):
Minor comment, otherwise LGTM!

PR #77, opened 2021-12-21
Implement post-editing

plausibly:
Issue #74 
- Update frontend API to have update post PUT request
- Clicking 'Edit' when viewing (own) post will open editing UI. This has input textfields with default value of what the post is.
- Can edit and change post title, body, location, capacity
- Some error handling when invalid capacity is entered, invalid lengths, etc..

image
image
image
image

**Updated UI:**
image


client/src/components/ViewPostDialog.tsx
```diff
@@ -115,7 +116,7 @@ function MoreOptions(props: {
       >
         {props.isAuth ? (
           <>
-            <MenuItem onClick={closeMenu}>Edit</MenuItem>
+            <MenuItem onClick={() => props.toggleEdit(true)}>Edit</MenuItem>
```
andy9775, line 119: I'm assuming `isAuth` specifies if the user created the post they are trying to delete/update? If so, lets update that variable name, seems like it's just checking of the user is logged in. Something like `userHasCreatedPost`

client/src/components/ViewPostDialog.tsx
```diff
@@ -206,6 +207,126 @@ function CapacityBar(props: { maxCapacity: number }) {
...
+      setMsg('Body must be atleast 25 characters');
+      showAlert(true);
+    } else if (form.title === '' || form.location === '') {
+      setMsg('Enter all required fields');
+      showAlert(true);
+    } else if (isNaN(form.capacity)) {
+      setMsg('Capacity must be a number');
+      showAlert(true);
+    } else {
+      console.log(form);
```
andy9775, line 242:
```suggestion

```

andy9775 (changes requested):
I'm curious why you didn't reuse the create post component? You should be able to update it to take the post state via props of the post being edited. From this, if `defaultState` is not undefined you could update the `create post` button text to `update post` and use the appropriate API endpoint. We've already tested all that logic, all you would need to do is write a test to ensure that when default state is passed in, the data is visible in the form fields, the text is correct for the button, and that the updatePost api endpoint is hit.

plausibly:
> I'm curious why you didn't reuse the create post component? You should be able to update it to take the post state via props of the post being edited. From this, if `defaultState` is not undefined you could update the `create post` button text to `update post` and use the appropriate API endpoint. We've already tested all that logic, all you would need to do is write a test to ensure that when default state is passed in, the data is visible in the form fields, the text is correct for the button, and that the updatePost api endpoint is hit.

I thought about that, but my logic was that it will look better to have textboxes in a similar format to the post view rather than opening another dialog on top of it

andy9775 (approved):

plausibly:
Requested re-review because I made slight changes after fixing merge conflicts. Just making sure I didn't break anything.

Also @andy9775, how do we want to handle editing capacity? I was thinking they cannot reduce capacity below the amount of people that have already checked in

andy9775:
Sounds good. Again, lets focus on getting the things we set out complete before getting deep into all the different edge cases - we could spend days on one feature, thinking of all the possible edge cases and so on.

andy9775 (approved):

PR #78, opened 2021-12-22
Enable file uploads to backblaze.

andy9775:
Enables file uploads to backblaze, uploaded files are public and available directly from backblaze.

closes: #50


andy9775:
Note, there aren't any tests here since this really is just connecting up backblaze and using their library

server/services/fileManager.ts
```diff
@@ -0,0 +1,34 @@
...
+      bucketId: process.env.BACKBLAZE_BUCKET_ID as string,
+    });
+    await this.b2.uploadFile({
+      fileName: fileName,
+      data: fs.readFileSync(localFilePath),
+      uploadUrl: uploadUrl.data.uploadUrl,
+      uploadAuthToken: uploadUrl.data.authorizationToken,
+    });
+
+    return `https://f004.backblazeb2.com/file/${process.env.BACKBLAZE_BUCKET_NAME}/${fileName}`;
```
andy9775, line 32: @plausibly Can you add the appropriate env variables to heroku please?

plausibly: Sure, just DM me what the variable should be equal to

server/services/fileManager.ts
```diff
@@ -0,0 +1,34 @@
+import b2 from 'backblaze-b2';
+import fs from 'fs';
+
+export default class FileManager {
+  protected b2: b2;
+
+  constructor() {
+    this.b2 = new b2({
+      applicationKeyId: process.env.BACKBLAZE_APP_KEY_ID as string,
```
plausibly, line 9: Can you update the post controller test to mock this class? It's failing because of 
```FAIL controllers/v1/tests/post.test.ts
  ● Test suite failed to run

    Invalid accountId or applicationKeyId
```

plausibly (changes requested):
Minor comment with the tests failing. Can you also integrate this in the frontend?

plausibly:
Should there also be some validation so that the only acceptable filetypes are images? (PNG, JPG/JPEG, ...)

server/services/fileManager.ts
andy9775: lets add the keys as part of the test process via secrets. The test doesn't hit backblaze anyway

andy9775:
> 

Sure, any specific types that you can think of? We don't want to be too restrictive.

server/services/fileManager.ts
andy9775: I used my work credentials to test it. Can you sign up for an account? It's also easier to have one person maintain all the credentials. You don't need to provide payment details.

plausibly:
> > 
> 
> Sure, any specific types that you can think of? We don't want to be too restrictive.

JPG, JPEG, PNG should be fine. We can update createform to specify that these are the only allowed types.

server/services/fileManager.ts
plausibly: Ok. Can you update the github actions to set the env variables from secrets? I'll setup the secrets on github after

andy9775:
> Minor comment with the tests failing. Can you also integrate this in the frontend?

The frontend is already uploading a file, its just not being read on the backend.

plausibly:
> The frontend is already uploading a file, its just not being read on the backend.

Kk. Can you deploy and test everything on the frontend to make sure it works?

plausibly (changes requested):
Update `ViewPostDialog.tsx` to load in the thumbnail instead of the placeholder image (see the TODO message). Then deploy locally and check if it works.

I added the secrets to the repo, but it is still failing tests.. You might have to add it to your own repo, I can DM you them tmr

plausibly (changes requested):
`PostPreview` needs to show the thumbnail instead of placeholder image.

`CreatePost` thumbnail upload function might need to be updated, tested locally and it doesn't work.

Can you deploy it locally and do some manual tests to make sure everything works? Try uploading a large image to see if its scaled down properly on dashboard cards. I can send you the backblaze info if you need it

client/src/components/ViewPostDialog.tsx
```diff
@@ -355,8 +355,7 @@ export default function ViewPostDialog(props: {
             }}
           >
             <img
-              // TODO src={props.postUser.thumbnail}
-              src='https://i.imgur.com/8EYKtwP.png'
+              src={props.postUser.thumbnail}
```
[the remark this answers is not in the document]
andy9775: ya, I did some local testing

plausibly (changes requested):
`PostPreview` needs to show the thumbnail instead of placeholder image.

Add validator to createpost to only allow and show image types in the file picker. Test both frontend and backend with docker deploy (use the ui not postman), it doesn't seem to work on my end. Maybe the format that CreatePost sends to the API is incorrect

client/src/components/PostPreview.tsx
```diff
@@ -65,13 +63,14 @@ export default function PostPreview(props: {
...
         }}
       >
-        <CardMedia
-          component='img'
-          //TODO: once we set this up src={props.postUser.thumbnail}
-          src='https://i.imgur.com/8EYKtwP.png'
-          alt='placeholder'
-          sx={{ maxWidth: '500px', maxHeight: '145px' }}
-        />
+        {!!props.postUser.thumbnail ? (
```
plausibly, line 66: `!!` is not really necessary

client/src/components/CreatePost.tsx
```diff
@@ -22,6 +22,16 @@ import { LocationPickerMap } from './LocationMap';
...
+  file: File | undefined;
+  tags: string[];
+  capacity: number;
+  location: string;
+  coords: { lat: number; lng: number };
+}
+
 function CreatePost() {
   const [openPopup, setOpenPopup] = useState(false); // for preview popup
   const [isAlertOpen, showAlert] = useState(false); // for snackbar
```
plausibly, line 37: Show an alert if they attempt to upload an invalid file type

plausibly (changes requested):
Can you adjust the style if a post has no thumbnail so that everything looks aligned:

image

Also check to make sure if uploading massive images (in terms of height/width) doesn't mess with the formatting of our preview cards, or the post dashboard.

client/src/components/PostPreview.tsx
andy9775: It's good practice

andy9775:
That's the default behaviour. Are we changing it in general, or conditionally when there is no image?
post view

client/src/components/CreatePost.tsx
andy9775: Do we have a standard alert component?

plausibly:
> That's the default behaviour. Are we changing it in general, or conditionally when there is no image? post view

Might be best to change it in general.

client/src/components/CreatePost.tsx
plausibly: There's something in `CreatePost` that handles error messages in the snackbar. I think `setMsg` and `showAlert` can be used here

andy9775:
> > That's the default behaviour. Are we changing it in general, or conditionally when there is no image? post view
> 
> Might be best to change it in general.

That should be a seperate PR in that case.

plausibly:
@andy9775 Can you adjust the thumbnail style when viewing a post so that it takes up the whole width of the row? For example, ,if I upload a large image, it's just scaled into a small square onto the post.

It should take up more space

image

andy9775:
If it takes up the width of the whole row, it'll just be stretched. e.g. if you have a 200x200 image but I force it to be 1000x200, it'll be off. I can do it, but it does't make a whole lot of sense to force it.

plausibly:
> If it takes up the width of the whole row, it'll just be stretched. e.g. if you have a 200x200 image but I force it to be 1000x200, it'll be off. I can do it, but it does't make a whole lot of sense to force it.

It can be scaled based on how big the image is. Like if the image is large, it should take up more of the width. Right now, all images are scaled down into a small box. I tested uploading a large landscape photo and it just turns into a square

client/src/components/CreatePost.tsx
andy9775: That would require a refactoring. We need to have a standard alert component shared across different pages. Leaving as a todo for now

andy9775:
The image size is lost, I'm sizing based on the viewport and letting the image fill its self in

nginx/conf.d/default.conf.template
```diff
@@ -2,6 +2,13 @@ server {
     listen       $PORT;
     server_name  localhost;
 
+    location /file/ {
+                proxy_pass https://f004.backblazeb2.com;
```
plausibly, line 6: Just curious, is it always f004?

server/services/fileManager.ts
```diff
@@ -0,0 +1,38 @@
...
+    });
+    await this.b2.uploadFile({
+      fileName: fileName,
+      data: fs.readFileSync(localFilePath),
+      uploadUrl: uploadUrl.data.uploadUrl,
+      uploadAuthToken: uploadUrl.data.authorizationToken,
+    });
+
+    return `${
+      process.env.ENV === 'production' ? 'https' : 'http'
```
plausibly, line 33: `PAGE_URL` on heroku already has `https://` and ends in a `/`.
 
i.e `PAGE_URL=https://university-board.herokuapp.com/` (we cannot change this or it will break code elsewhere)

Deploying this on heroku, thumbnails have the URL `http://https:/university-board.herokuapp.com//file/uoftbored/4f814bf9-ca55-480e-8260-59470a26527b.jpg`

No need to put `/` before file, or the `http` prefix.

client/src/components/ViewPostDialog.tsx
```diff
@@ -354,13 +354,13 @@ export default function ViewPostDialog(props: {
...
-              src='https://i.imgur.com/8EYKtwP.png'
-              alt='Thumbnail'
-              height='400px'
-              width='400px'
-            />
+            {!!props.postUser.thumbnail ? (
+              <img
+                src={props.postUser.thumbnail}
+                alt='Thumbnail'
+                style={{ maxHeight: '60vw', maxWidth: '60vw' }}
```
plausibly, line 361: 60vw seems too big, I think `30vw` for maxHeight is better. Can you also add padding to the bottom of the image, sometimes the image looks too close to the body text. Just  `pb: 1` should be fine

plausibly (changes requested):
Looks good for the most part, just some minor issues, I also noticed it takes a few seconds after loading the page before the thumbnails are fetched, is this due to the NGINX proxying?

andy9775:
> Looks good for the most part, just some minor issues, I also noticed it takes a few seconds after loading the page before the thumbnails are fetched, is this due to the NGINX proxying?

It depends on the size of the image. They're not cached in a CDN anywhere so a large image will take some time to hit heroku and get back to you

nginx/conf.d/default.conf.template
andy9775: Yes

plausibly (approved):
LGTM

PR #79, opened 2021-12-22
Fix mailing issue

plausibly:
uoft seems to be blocking embedded hyperlinks, so the emails now require manually copy and pasting a link


andy9775 (approved):
LGTM!

PR #82, opened 2021-12-24
Enable user checkin functionality

andy9775:
Allows user to checkin/out of an event


client/src/components/ViewPostDialog.tsx
```diff
@@ -159,49 +159,54 @@ function LikeButton(props: { numLikes: number }) {
...
+      props.isUserCheckedIn = false;
+      props.usersCheckedIn -= 1;
     } else {
-      setCapacity((prev) => (prev > 0 ? prev - 1 : prev));
+      const result = await api.checkin(props.postID);
+      if (result.status !== 409) {
+        props.isUserCheckedIn = true;
+      }
+      // TODO indicate a standard alert to the user that the event could not be
+      // checked into (over capacity)
```
plausibly, line 184: shouldn't this be done in this PR since its a capacity related error?

client/src/components/ViewPostDialog.tsx
```diff
@@ -159,49 +159,54 @@ function LikeButton(props: { numLikes: number }) {
...
-        <Button onClick={handleCheckIn} variant='outlined'>
-          Check In
-        </Button>
-      )
-    ) : (
-      <Button disabled variant='outlined'>
-        AT CAPACITY
-      </Button>
-    );
+  const buttonHandler = props.isUserCheckedIn ? (
```
plausibly, line 188: Add some tests to make sure this behaves correctly. You can check `PostDashboard.test.tsx` and do a similar method of mocking post responses.

plausibly (approved):
Looks good, just some minor comments. Deploy locally and use multiple accounts to make sure nothing is buggy, try checking in at the same time for an event with 1 capacity.

client/src/components/ViewPostDialog.tsx
andy9775: As mentioned on the other pr, we need a standard alert component to be shared across the app. Account confirmation info box is different from the one on the post creation. I'm leaving messages as TODO's since it will change either way.

client/src/components/ViewPostDialog.tsx
andy9775: I think for something like this its fine as is.

plausibly:
I think we should have a route that shows all of the events that the current user checked into,  this can be added as an option in the menu under my posts.

andy9775:
Sounds good. I think we should focus on getting the baseline features in though before feature-creeping.

plausibly (changes requested):
Same issue as #84 when deployed to heroku

plausibly (changes requested):
Same issue
image

plausibly (approved):

danielduong8 (approved):

danielduong8 (dismissed):
LGTM

danielduong8 (approved):

plausibly (commented):
Can you add a useEffect in `viewPostDialog` to fetch data after the user hits checkin? The checkin button takes 5 seconds before it updates the checkin status which makes it look very buggy. Looks good otherwise.

`getPosts` - can you make sure it also returns total number of users checked in, alongside with `capacity`.  Update `PostUserPreview` to reflect these changes (capacity, and `totalCheckedIn`)

andy9775:
We should have used context or redux for state management - useEffect is a bit hacky in this context. Since it affects checking and report I've opted to increase the frequency of polling rather than have multiple useEffect's

plausibly (approved):

PR #84, opened 2021-12-24
Enable post reporting.

andy9775:
A user can report a post which is recorded

### DIFF
--- a/client/src/components/ViewPostDialog.tsx
+++ b/client/src/components/ViewPostDialog.tsx
@@ -36,8 +36,6 @@ function MoreOptions(props: {
   userHasCreatedPost: boolean;
   toggleEdit: React.Dispatch<React.SetStateAction<boolean>>;
   useNavigate: NavigateFunction;
-  isAuth: boolean;
-  closeDialog: Function;
   didUserReport: string;
 }) {
   const [isOpen, toggleMenu] = React.useState(false);
@@ -121,7 +119,7 @@ function MoreOptions(props: {
         ) : (
           <></>
         )}
-        {!props.isAuth ? (
+        {!props.userHasCreatedPost ? (
           props.didUserReport === '0' ? (
             <MenuItem onClick={reportPost}>Report</MenuItem>
           ) : (
--- a/client/src/components/ViewPostDialog.tsx
+++ b/client/src/components/ViewPostDialog.tsx
@@ -118,11 +118,13 @@ function MoreOptions(props: {
         ) : (
           <></>
         )}
-        {!props.didUserReport ? (
-          <MenuItem onClick={reportPost}>Report</MenuItem>
-        ) : (
-          <MenuItem disabled>Reported</MenuItem>
-        )}
+        {!props.isAuth ? (
+          !props.didUserReport ? (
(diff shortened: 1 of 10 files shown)


client/src/components/ViewPostDialog.tsx
```diff
@@ -123,7 +126,9 @@ function MoreOptions(props: {
         ) : (
           <></>
         )}
-        <MenuItem onClick={reportPost}>Report</MenuItem>
+        {!props.didUserReport ? (
```
plausibly, line 129: Just a nit, I don't think we should hide it if the post is already reported -- maybe just say "Post has already been reported". On other sites, if you report a post, the button is not hidden.

server/services/emailService.ts
```diff
@@ -48,6 +48,7 @@ export default class EmailService {
     To continue with your account registration, please confirm your email address by visiting: 
     
     ${confirmURL}`;
+    console.log(body);
```
plausibly, line 51: Do we want to log this?

andy9775: I was logging it out to test locally - easier than going through the whole signup flow.

client/src/components/ViewPostDialog.tsx
```diff
@@ -71,15 +72,17 @@ function MoreOptions(props: {
...
-    // TODO need to fix spam / backend interactions
-
     api
       .reportPost(props.postID)
       .then((res) => {
+        props.didUserReport = true;
         if (res.status === 204) {
           setMsg('Post has been reported.');
+        } else if (res.status === 200) {
+          setMsg('Post has been reported and deleted.');
```
plausibly, line 82: After you fix the merge conflicts with routing changes, can you redirect to dashboard? Or navigate(-1)? Since the page will no longer exist

client/src/components/ViewPostDialog.tsx
```diff
@@ -123,7 +127,11 @@ function MoreOptions(props: {
         ) : (
           <></>
         )}
-        <MenuItem onClick={reportPost}>Report</MenuItem>
+        {!props.didUserReport ? (
+          <MenuItem onClick={reportPost}>Report</MenuItem>
+        ) : (
+          <MenuItem disabled>Reported</MenuItem>
```
plausibly, line 133: Should not show this option at all if the user authored the post

plausibly (changes requested):
Add some tests for the frontend.  When I deployed on Heroku it was showing "Reported" for posts that weren't reported
image

plausibly:
I checked the response from getPost, it seems like didUserReport is returning a number instead of boolean. Might need to update the boolean logic on frontend to fix the "Reported" issue

image

andy9775:
> Add some tests for the frontend. When I deployed on Heroku it was showing "Reported" for posts that weren't reported image

If it's a heroku only issue, tests won't address the issue.

plausibly (approved):
LGTM

danielduong8 (approved):

plausibly (approved):
Same thing with check-in, since it takes 5s to fetch post data, there is a delay from when the user clicks "Report" to when it says "Reported". Might want to make it fetch data when the user clicks report. Other than that, LGTM

PR #86, opened 2021-12-28
Limit the maximum number of failed login attempts for a user.

andy9775:
After a specified number of failed login attempts, the user is locked out for a defined period of
time. Once the time has passed, the user can attempt to login again.

closes: 12

### DIFF
--- a/server/controllers/v1/tests/user.test.ts
+++ b/server/controllers/v1/tests/user.test.ts
@@ -1,9 +1,11 @@
 import argon2 from 'argon2';
+import mockdate from 'mockdate';
 import db from '../../../models';
 import {
   dbSync,
   makeUser,
   makeUserWithPass,
+  makeValidUser,
 } from '../../../models/tests/testHelpers';
 import { User } from '../../../models/user';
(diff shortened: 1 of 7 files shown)


andy9775:
Note, leaving out f/e changes as we need to standardize the error component

plausibly:
Just took a quick look, I don't think `mockdate` is necessary. Jest allows you to mock the current time, I did something similar in the postdashboard test
https://jestjs.io/docs/jest-object#jestsetsystemtimenow-number--date
  ```
jest.useFakeTimers('modern')
jest.setSystemTime(new Date(...) )
```

**Review by danielduong8 — APPROVED** [review]
LGTM

File: server/controllers/v1/tests/user.test.ts [hunk 1]
```diff
@@ -143,6 +145,71 @@ describe('v1 - User Controller', () => {
       test('Password should not be included in serialized result', () => {
         expect(signInResponse.data.result.password).not.toBeDefined();
       });
+
```
plausibly, line 147: Add test: Other users should not be affected if someone is locked out.

(User 1 is locked out, user 2 can still login normally)

PR #87, opened 2021-12-28
Create 404 page for nginx

andy9775:
closes: 40
Screen Shot 2021-12-28 at 3 38 36 PM

### DIFF
--- a/nginx/html/404.html
+++ b/nginx/html/404.html
@@ -1,0 +1,64 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8" />
+    <link rel="icon" href="/favicon.ico" />
+    <meta name="viewport" content="width=device-width, initial-scale=1" />
+    <meta name="theme-color" content="#000000" />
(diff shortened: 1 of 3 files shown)


danielduong8 (approved):
Just a minor comment... don't know if this was what you were going for but it looks like the header at the top is supposed to sort of recreate our header from the dashboard? I think it might be fine to just have the UBoard text somewhere without the blue bar. Otherwise, LGTM!

plausibly (approved):

nginx/html/404.html
```diff
@@ -0,0 +1,64 @@
...
+        padding-top: 15vh;
+      }
+      .error {
+        font-size: 6rem;
+      }
+      .msg {
+        font-size: 1.25rem;
+      }
+    </style>
+    <title>UofTBoard</title>
```
plausibly, line 52: Clicking this should return to /dashboard

andy9775:
Ya, I wanted to keep it consistent. That way it doesn't look like just a general 404 page.

PR #97, opened 2021-12-31
Add post type variations

### LINKED ISSUE #90: Add post type variations
Original poster (OP) should have an option to select from various post types, each with different forms for the OP and user to fill out.

There should be post type variations such as 'Textbooks', 'Roommates', 'Applications', etc...

OPs and non-OPs should have different views for the same post.

OP should be able to set up call-to-action (CTA) buttons and users should be able to click on them.

Users should be able to filter content through a dropdown which selects from the types of posts.

danielduong8:
Close #90
- Add dropdown to select post type in post creation
  - Form only appears after type is selected
  - Button labels change dynamically based on the type of post selected
  - Capacity labels change dynamically based on the type selected
  - Dynamically adjust form for post editing
- Add dropdown on dashboard to filter between different types of posts
- Show type of post on preview
- Fix bad length truncation and update tests


server/controllers/v1/post.ts
```diff
@@ -620,6 +638,7 @@ export default class PostController {
    */
   async createPost(
     userID: string,
+    type?: string,
```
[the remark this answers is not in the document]
danielduong8: checked

plausibly:
Note: Whoever merges first will need to update the MapView to only show for 'Events'. Haven't made that PR yet, but just remind me to look into this if I forget.

client/src/components/PreviewPopUp.tsx
```diff
@@ -66,9 +70,15 @@ export default function PreviewPopUp(props: {
...
             <Grid container sx={{ py: 1 }}>
               <Grid item xs={12}>
-                <Typography>Capacity: {eventCapacity}</Typography>
+                <Typography>
+                  {typeLabels[props.type]}: {eventCapacity}
+                </Typography>
               </Grid>
-              <Grid item xs={12}>
+              <Grid
+                display={type === 'Events' ? undefined : 'none'}
```
plausibly, line 78: Did you fix this from overlapping on tags?

client/src/components/Tags.tsx
```diff
@@ -9,7 +9,7 @@ import Typography from '@mui/material/Typography';
 /* Generate pre-formatted TagItems based on an array of provided tags */
 export default function GenerateTags(props: { tags: string[], space?: number }) {
   return (
-    <Stack direction='row' spacing={props.space ? props.space : 1} style={{ alignItems: 'center' }}>
```
plausibly, line 12: Change this back and just don't pass space props so it can default to 1. I think I did this because I needed custom spacing somewhere, although i forgot where

plausibly (changes requested):
See comments. Did you also fix the type overlapping with tags?

client/src/components/Tags.tsx
```diff
@@ -18,9 +18,10 @@ export default function GenerateTags(props: { tags: string[], space?: number })
 }
 
 export function Tag(props: { tag: string }) {
+  const truncated = props.tag.length > 10 ? props.tag.slice(0, 7) + '...' : props.tag;
```
andy9775, line 21: Isn't the limit 3 tags?

client/src/containers/PostDashboard.tsx
```diff
@@ -9,16 +9,31 @@ import Header from '../components/Header';
...
 
+const postTypes = [
+  'All',
+  'Events',
+  'Clubs',
+  'Textbooks',
+  'Roommates',
+  'Opportunities',
+];
+
```
andy9775, line 27: The types should really only exist in one spot. `constants.ts`? Helps with misspellings and such

andy9775 (approved):
Just a few minor comments. otherwise LGTM

client/src/components/Tags.tsx
danielduong8: This is to limit the length of an individual tag which is a string, since we don't want them to be too long and overflow on a post preview.

plausibly: @danielduong8 don't truncate the tag in the post dialog, only truncate in preview. Perhaps add a prop `shouldTruncate` to help with this

plausibly, line 21: This causes the tag to also be truncated in `ViewPostDialog` which looks weird. Only truncate on the previews

plausibly (changes requested):

plausibly (approved):
Looks good. Can you deploy and test everything, including with google maps API, to make sure nothing broke

danielduong8:
> Looks good. Can you deploy and test everything, including with google maps API, to make sure nothing broke

we should merge fix/polling first before we do any api key deploys

PR #102, opened 2021-12-31
Implement map view on dashboard

plausibly:
Ability to toggle 'map view' which shows all in-person events on a map. The marker is color-coded based on capacity, where green is basically an empty event, medium is partially filled, and red is at capacity.

Has a manual 'refresh' button so that the polling doesn't interrupt the users interaction with the map.

I had to style the map with media queries as it required explicitly defining h/w. It works, but sometimes the map wont be fully centered depending on screen size.... oh well lol

#97 'Toggle map view' should only be visible for `Events`

#107 adjust page url after merged

image
image
image


client/src/components/LocationMap.tsx
```diff
@@ -138,6 +147,160 @@ export function LocationPickerMap(props: {
...
+          <p> Located @ ${curr.location} </p> 
+          ${hasAttendance}
+          <a href="/${curr.id}"> Read More </a>
+          </div>`,
+        });
+
+        const percentFilled =
+          (curr.capacity > 0 ? (curr.usersCheckedIn / curr.capacity) : 1) * 100;
+          
+        // red marker indicates event is full, blue is almost filled, green is almost empty event (less than 50%)
```
andy9775, line 243: Shouldn't the marker colours relate to the type of the event? If I'm looking for events I'm not looking for just any event which has capacity, I'd be looking for something specific (type).

Just a thought.

andy9775 (approved):
Just some minor comments, otherwise LGTM

client/src/components/LocationMap.tsx
plausibly: The color coding based on capacity was suggested by Ilir.  We could change the icon itself (while maintaining different color codes) for the type of event?

andy9775: Ya, different icons make sense as well + colours.

danielduong8: this is fine, only posts of type 'event' has locations for now

danielduong8:
may need to fix media queries, the map overflows for certain mobile devices, try the ipads for example

PR #104, opened 2022-01-01
Add error handling to `fileManager`

plausibly:
`fileManager` now checks if backblaze failed. Without an error check, our entire site will break if the line `this.b2 = new b2({..})` raises an error.


andy9775 (changes requested):
Realistically the application should fail to start-up if this happens, or there should be a fallback, or we should inform the user that uploading files isn't supported. This is kind of a no-op at this point.

plausibly:
> Realistically the application should fail to start-up if this happens, or there should be a fallback, or we should inform the user that uploading files isn't supported. This is kind of a no-op at this point.

I'll add an error message to create post. I just want to make sure the site doesn't go down if something happens with backblaze.

andy9775 (approved):
How do we handle the case of backblaze coming back up?

This only catches the issue if backblaze is down at the same time that the app is starting up. What happens if it goes down while the app is up? What happens if it comes back up? If backblaze is down during startup, it will only be re-detected if the app is restarted. 

Approved, but makes no sense.

PR #105, opened 2022-01-01
Fix polling and interactions

danielduong8:
- Increase polling interval to 3s
- Fix dependency warnings by using Callbacks
- Fix blocking on interactions in b/e by sending empty responses
- Fetch new post data immediately on interaction

### DIFF
--- a/client/src/components/ViewPostDialog.tsx
+++ b/client/src/components/ViewPostDialog.tsx
@@ -139,7 +139,6 @@ function MoreOptions(props: {
 
 /* Like button. Handles liking/unliking a post */
 function LikeButton(props: {
-  interactionBit: boolean;
   setInteractionBit: React.Dispatch<React.SetStateAction<boolean>>;
   numLikes: number;
   doesUserLike: string;
@@ -154,7 +153,7 @@ function LikeButton(props: {
     } else {
       await api.unlikePost(props.id);
     }
-    props.setInteractionBit(!props.interactionBit);
+    props.setInteractionBit((bit) => !bit);
   };
 
   const likeButton = isLiked ? (
@@ -184,7 +183,6 @@ function LikeButton(props: {
 }
 
 function CapacityBar(props: {
-  interactionBit: boolean;
   setInteractionBit: React.Dispatch<React.SetStateAction<boolean>>;
   maxCapacity: number;
   postID: string;
@@ -203,7 +201,7 @@ function CapacityBar(props: {
         // checked into (over capacity)
       }
     }
-    props.setInteractionBit(!props.interactionBit);
+    props.setInteractionBit((bit) => !bit);
   };
 
   const buttonHandler =
@@ -476,7 +474,7 @@ export default function ViewPostDialog() {
   }, [postid, userContext.data]);
 
   React.useEffect(() => {
-    if (!error) {
+    if (!error && !isEditing) {
       fetchData();
     }
   }, [fetchData, error, isEditing, interactionBit]);
@@ -583,7 +581,6 @@ export default function ViewPostDialog() {
             <Stack direction='row' sx={{ px: 4, pb: 5 }}>
               {Number(postData.capacity) > 0 ? (
                 <CapacityBar
-                  interactionBit={interactionBit}
                   setInteractionBit={setInteractionBit}
                   maxCapacity={Number(postData.capacity)}
                   postID={postData.id}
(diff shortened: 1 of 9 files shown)


server/routes/v1/post.ts
```diff
@@ -122,7 +122,7 @@ postRouter.put('/:postid/upvote', async (req: Request, res: Response) => {
       getAuthUser(res).id,
       req.params.postid
     );
-    res.status(result.status);
+    res.status(result.status).json();
```
plausibly, line 125: Why json? Isn't this a 204? For the error case we can return the json and status code. same for the other ones

server/routes/v1/post.ts
```diff
@@ -159,7 +159,7 @@ postRouter.put('/:postid/checkout', async (req: Request, res: Response) => {
...
```
plausibly, line 162: same

client/src/components/ViewPostDialog.tsx
```diff
@@ -472,31 +471,14 @@ export default function ViewPostDialog() {
...
-  }, [isEditing]);
-
-  React.useEffect(() => {
-    /* Fetch incase data has changed / post was edited */
     if (!error) {
-      const interval = setInterval(() => {
-        fetchData();
-      }, 500);
-      return () => clearInterval(interval);
+      fetchData();
```
plausibly, line 478: Do not fetch data if the user `isEditing`

plausibly (changes requested):

server/routes/v1/post.ts
andy9775: Agreed. This shouldn't be needed.

andy9775 (changes requested):

client/src/components/ViewPostDialog.tsx
danielduong8: its fine, if anything it will ensure that they get the most recent update to the post before they make changes to it

server/routes/v1/post.ts
danielduong8: the reason for this was mentioned in my commit comment (hit the triple dots)

the website was blocking or stalling on any interaction because despite the 204 status, they would still be waiting for a response, therefore, sending empty .json() resolves this issue

server/routes/v1/post.ts
danielduong8: see response above

danielduong8:
the client will stall and show an error in the console unless .json() or .send() or something is sent, doesn't seem to be a way around it

i could however change them to .end() if thats better

client/src/components/ViewPostDialog.tsx
plausibly: This might cause weird behavior because the input boxes have default value of what the postdata is, but if we keep fetching data, it might bug it out. Also, if we fetch a post at the same time that the user clicks 'update' (after finishing editing) there will be lag before the request is sent.

client/src/components/ViewPostDialog.tsx
```diff
@@ -139,19 +139,22 @@ function MoreOptions(props: {
...
-  const isLiked = props.doesUserLike == '1';
+  const isLiked = props.doesUserLike === '1';
 
   const handleClick = async () => {
     if (!isLiked) {
       await api.likePost(props.id);
     } else {
       await api.unlikePost(props.id);
     }
+    props.setInteractionBit(!props.interactionBit);
```
plausibly, line 157: If you just want to set it to the opposite value ,you can just do `props.setInteractionBit((prev) => !prev)`.. you dont need `props.interactionBit`

plausibly, line 157: If you just want to set it to the opposite value ,you can just do `props.setInteractionBit((prev) => !prev)`.. you dont need `props.interactionBit`

client/src/components/ViewPostDialog.tsx
```diff
@@ -472,31 +473,13 @@ export default function ViewPostDialog() {
...
-  }, [isEditing]);
-
-  React.useEffect(() => {
-    /* Fetch incase data has changed / post was edited */
     if (!error) {
-      const interval = setInterval(() => {
-        fetchData();
-      }, 500);
-      return () => clearInterval(interval);
+      fetchData();
```
plausibly, line 480: don't fetch on isEditing :(

plausibly (changes requested):

client/src/containers/PostDashboard.tsx
```diff
@@ -28,7 +29,7 @@ function RecentPosts(props: {
   const [openedPost, setOpenedPost] = React.useState(false);
 
   const checkForPosts = React.useCallback(() => {
-    if (!openedPost) {
+    if (!openedPost && !props.openedCreate) {
```
plausibly: Remove `openedPost` I think its useless now that we have removed post dialog and replaced it with a router

plausibly (approved):
LGTM, just minor comment. Test on heroku as well

PR #106, opened 2022-01-01
Fix 404 on invalid post

plausibly:
Our backend would not return anything if an error was caught, updated this to return 500 error. This would also cause bugs for trying to visit posts that don't exist, since the API would not return anything it will just get stuck on 'loading'.

note to self: deploy on heroku and test later

### DIFF
--- a/client/src/App.tsx
+++ b/client/src/App.tsx
@@ -15,6 +15,7 @@ import './App.css';
 import { User } from 'models/user';
 
 import ServerApi from './api/v1';
+import ViewPostDialog from './components/ViewPostDialog';
 
 const api = new ServerApi();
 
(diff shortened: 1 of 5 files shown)


server/routes/v1/post.ts
```diff
@@ -27,6 +27,7 @@ postRouter.get('', async (req: Request, res: Response) => {
     return res.status(result.status).json(result);
   } catch (err) {
     console.error(err);
+    res.status(500);
```
andy9775, line 30: [Already sets a status message](https://github.com/GDSCUTM-CommunityProjects/UBoard/blob/main/server/middleware/auth.ts#L23). Failing to get the user shouldn't be a 500.

Here and elsewhere

andy9775 (changes requested):

server/routes/v1/post.ts
plausibly: Which other ones? I see the getPost route also contains getauthuser, but we still need a fallback error code. Going to the route with an invalid postid will not return a status, so the frontend just gets stuck waiting for a response.

PR #107, opened 2022-01-02
Update `getPost` to properly return `404` on error

plausibly:
Fetching an invalid post, or one that was previously deleted, would not properly return `404` error and instead just timeout after a few mins. I fixed this by wrapping the DB query in a try catch.

On the frontend page URL was adjusted to be `post-postid` instead of `/postid` (I wanted to do `/posts/id` but this was causing issues with axios).

After these changes, going to an invalid post, or viewing a post that gets deleted, will redirect to the `/404` error page.

### DIFF
--- a/client/src/App.tsx
+++ b/client/src/App.tsx
@@ -93,7 +93,7 @@ function App() {
         <Routes>
           <Route path='/' element={<AuthContainer />} />
           <Route path='/dashboard' element={<ProtectedRoute destination={<PostDashboard />} />} />
-          <Route path='post-:postid' element={<ProtectedRoute destination={<ViewPostDialog />} />}  /> 
+          <Route path='/post-:postid' element={<ProtectedRoute destination={<ViewPostDialog />} />}  /> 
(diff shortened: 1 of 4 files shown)


client/src/App.tsx
```diff
@@ -93,7 +93,7 @@ function App() {
         <Routes>
           <Route path='/' element={<AuthContainer />} />
           <Route path='/dashboard' element={<ProtectedRoute destination={<PostDashboard />} />} />
-          <Route path=':postid' element={<ProtectedRoute destination={<ViewPostDialog />} />}  /> 
+          <Route path='post-:postid' element={<ProtectedRoute destination={<ViewPostDialog />} />}  />
```
andy9775, line 96:
```suggestion
          <Route path='/post-:postid' element={<ProtectedRoute destination={<ViewPostDialog />} />}  /> 
```

andy9775 (approved):

danielduong8 (approved):

PR #109, opened 2022-01-03
General fixes

daniduong:


client/src/components/LocationMap.tsx
```diff
@@ -268,9 +256,8 @@ export function EventsMapView(props: { posts: PostUserPreview[] }) {
       <Paper
         elevation={5}
         style={{
-          height: mapHeight,
-          width: mapWidth,
-          justifyContent: 'center',
+          height: '56vw',
```
plausibly, line 259: screenshot? how does it look now

plausibly (approved):

client/src/components/LocationMap.tsx
danielduong8: test resizing the window and different devices on university-board, it is currently deployed

plausibly: it is clipping on my screen

plausibly, line 259: undo this change, it seems to break on certain screens

plausibly (changes requested):

plausibly (approved):

